PR #26, opened 2015-01-08
Identify content items, organisations for each problem report

benilovj:
In order to present relevant anonymous feedback to users in departments and ALBs, we need to know which organisation is interested in a particular piece of feedback. This can be done by looking up at the orgs that relate to the piece of content that originated the problem report.

The org information is retrieved from the Content API and persisted; this saves having to fetch a list of all content relating to a particular org, and trying to get the feedback for each one of those paths.

More details about the implementation in the commit messages.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,0 +1,3 @@
+class ContentItem < ActiveRecord::Base
+  validates :path, presence: true
+end
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,3 +1,11 @@
 class ContentItem < ActiveRecord::Base
+  has_and_belongs_to_many :organisations
   validates :path, presence: true
+
+  before_create :fetch_organisations
+
+  private
+  def fetch_organisations
+    self.organisations = Organisation.for_path(path)
+  end
 end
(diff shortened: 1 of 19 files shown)


app/models/content_item.rb
```diff
@@ -0,0 +1,15 @@
...
+  has_and_belongs_to_many :organisations
+
+  validates :path, presence: true
+
+  before_create :fetch_organisations
+
+  private
+  def fetch_organisations
+    orgs_data = SupportApi::enhanced_content_api.organisations_for(path)
+    self.organisations = orgs_data.map {|org_info| Organisation.where(org_info).first_or_create! }
```
danielroseman: I don't think the organisation create belongs here: seems like that should happen in the Organisation class itself. Perhaps move all this logic into a `for_path` classmethod on Organisation.

lib/content_api/base_info_lookup.rb
```diff
@@ -0,0 +1,36 @@
+module ContentAPI
+  class BaseInfoLookup
+    def initialize(content_api)
+      @content_api = content_api
+    end
+
+    def applies?(path)
+      true
```
danielroseman: shouldn't this default to false?

app/models/content_item.rb
benilovj: fixed and force-pushed.

lib/content_api/base_info_lookup.rb
benilovj: fixed and force-pushed.

PR #32, opened 2015-03-02
Filter problem reports by organisation, day/month

benilovj:
```
In order to improve GOV.UK content which directly relates to my organisation or my specialist area,
As a publisher to GOV.UK
I want a list of only the recent feedback, and only relating to my org.
```

This PR introduces:

* an endpoint which returns anonymous problem reports. This endpoint allows:
    * filtering by content linked to a particular organisation
    * filtering by the day or month of the feedback
    * returning JSON (for machine consumption) or CSV
* an endpoint which lists all orgs that have ever received feedback (so that these can be iterated through)
* a rake task that can create CSVs containing the previous month's feedback, for every org that has received feedback in that month. These CSVs can then be sent to interested publishers within the organisations. This export is a stop-gap measure until a self-serve export can be added to FeedEx (which will reuse the endpoints added here).


app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -16,6 +18,24 @@ def totals
       end
     end
 
+    def index
+      if interval.nil?
+        head 422
+      elsif selected_organisation.nil?
+        head 404
+      else
+        @results = selected_organisation.problem_reports.where(created_at: interval)
```
evilstreak: I think it would be nice to make this `where` clause into a scope

config/routes.rb
```diff
@@ -21,6 +21,15 @@
     get '/problem-reports/:date/totals',
         constraints: { date: /\d{4}-\d{2}-\d{2}/ },
         to: 'problem_reports#totals'
+
+    get '/problem-reports/:period',
```
evilstreak: It seems a little odd to put `period` into the URL, but have `organisation_slug` as a query string, and then 404 if it's not provided.

I think it would be OK to use a query string if `organisation_slug` was optional, but if it's required let's find a way to put it into the URL.

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -16,6 +18,24 @@ def totals
       end
     end
 
+    def index
+      if interval.nil?
+        head 422
```
evilstreak: I think `422` is the correct status to use here, but it's worth pointing out that `totals` uses a `404` for a similar input validation.

lib/tasks/problem_reports_reports.rake
```diff
@@ -0,0 +1,25 @@
...
+    org_slugs = support_api.organisations_with_feedback.map {|org_hash| org_hash["slug"]}
+    org_slugs.each do |slug|
+      response = support_api.problem_reports_for(
+        organisation_slug: slug,
+        month: last_month,
+        format: :csv
+      )
+      next if response.code == 204
+
+      File.open("#{Rails.root}/tmp/#{slug}.problem-reports.#{last_month.strftime('%Y-%m')}.csv", "w") do |f|
```
evilstreak: Would it be better to output this to STDOUT and let the caller decide where to redirect the output to?

evilstreak:
A couple of comments about the rake task and the construction of the URL for the new report, but otherwise good

config/routes.rb
benilovj: I've made `organisation_slug` optional in  58b0901

app/controllers/anonymous_feedback/problem_reports_controller.rb
benilovj: I've made `#totals` and `index` consistent in this respect with 58b0901

lib/tasks/problem_reports_reports.rake
benilovj: This process will produce 100+ CSV files, one for each organisation, and each file will go to a separate org.

benilovj:
@evilstreak I've converted the rake task into a `bin` script in  d24bc21, as discussed in meatspace

benilovj:
I think I've now addressed all the comments.

evilstreak:
:+1:

PR #33, opened 2015-03-13
Remove low-quality single-word feedback from the Contacts DB

benilovj:
This feedback has been analysed and shown to be almost never actionable.


benilovj:
https://github.gds/gds/alphagov-deployment/pull/866 has been deployed. /cc @jennyd

PR #36, opened 2015-04-29
Bugfix: fixed NoMethod Error bugs

binaryberry:
The support-api was raising regular exceptions in
Errbit:`NoMethodError: undefined method 'content_item_path' for nil:NilClass`

Wrote a test to recreate the bug, fixed it by adding a default org lookup
which will assign to GDS any content items that could not be associated with
another organisation.

cc @benilovj
[Taiga story](https://tree.taiga.io/project/core-improvement-theme/us/6)

### DIFF
--- a/lib/content_api/default_org_content_lookup.rb
+++ b/lib/content_api/default_org_content_lookup.rb
@@ -1,0 +1,21 @@
+require 'content_api/base_info_lookup'
+
+module ContentAPI
+  class DefaultOrgContentLookup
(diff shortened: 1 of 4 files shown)


boffbowsh:
Makes sense to me, but you might also want a thumb from someone more familiar with this repo :+1:

benilovj:
:+1:

benilovj:
tested on preview, works as expected.

PR #37, opened 2015-04-30
Bugfix: path can no longer be empty string

binaryberry:
Tightened path validation, as Errbit was throwing error
"No implicit conversion of nil into String" when path was "".

Path can no longer be nil

[Taiga story](https://tree.taiga.io/project/core-improvement-theme/us/14)

cc @benilovj


benilovj:
:+1:

PR #38, opened 2015-04-30
Database migration to remove nil path entries

binaryberry:
Following a bug fix, entries with nil path will now be filtered by the validation
of the AnonymousContact class. Now replacing value of historic entries that
had a path set to nil, to a path of '/'.

cc @benilovj 

[Taiga](https://tree.taiga.io/project/core-improvement-theme/us/14)

Don't merge until #37 is deployed.


benilovj:
:+1:, will wait until #37 is out to merge.

PR #41, opened 2015-04-30
Fixes to environment-specific timezone issues

benilovj:
Fixes to some timezone and environment-specific issues around reporting to the Performance Platform that I discovered while running the test suite outside the VM (see details in the commit comments).

### DIFF
--- a/app/workers/problem_report_daily_totals_pp_uploader_worker.rb
+++ b/app/workers/problem_report_daily_totals_pp_uploader_worker.rb
@@ -15,7 +15,7 @@ class ProblemReportDailyTotalsPPUploaderWorker
     )
     support_api = GdsApi::SupportApi.new(Plek.find('support-api'))
 
-    date = Date.new(year, month, day)
+    date = Time.utc(year, month, day)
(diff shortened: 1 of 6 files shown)


app/workers/problem_report_daily_totals_pp_uploader_worker.rb
```diff
@@ -34,7 +34,7 @@ def transform_daily_problem_report_totals(date, totals)
     totals["data"].map do |entry|
       {
         "_id" => "#{date.to_time.strftime("%Y-%m-%d")}_#{entry["path"].gsub("/", "")}",
-        "_timestamp" => date.to_time.iso8601,
+        "_timestamp" => date.to_time.utc.strftime("%FT%T%:z"),
```
evilstreak: If the aim is to have `Date.new(year, month, day)` treated as midnight UTC, then would it be better to use the `#to_datetime` method and stick with the `#iso8601` call?

benilovj: Good point. Fixed in both places and force-pushed.

PR #42, opened 2015-05-05
Set path column to not null

binaryberry:
Migration to ensure no more path nil values make it into the database
following bugfix #37.

cc @benilovj 

[Taiga](https://tree.taiga.io/project/core-improvement-theme/us/14)


benilovj:
:+1:

PR #44, opened 2015-05-08
Migrate the corporate content PP upload from Support app

benilovj:
The monthly upload of corporate content stats to the PP used to live
and run within the Support app, but belongs within the Support API,
closer to the data.

This upload was originally added to the Support app in https://github.com/alphagov/support/pull/156.

This data supports provides data to the feedback modules of the
performance platform GOV.UK web traffic dashboards
(https://www.gov.uk/performance/web-traffic).


mikejustdoit:
Do this and https://github.com/alphagov/support/pull/233 need deploying together?

benilovj:
> Do this and alphagov/support#233 need deploying together?

The upload runs on the 1st of every month, so the timing is very forgiving.

PR #46, opened 2015-05-13
Derive the AnonymousContact url from path

benilovj:
Some time ago, the `AnonymousContact#path` started being used instead
of `AnonymousContact#url`.

Currently:

- no endpoints accept the `url` as an attribute
- `path` has a not-null constraint 
- all the non-null paths in the database have been cleaned up

This means that we can drop all remaining uses of the `url` field and derive it from `path` completely. The next step would be dropping the column from the database, once the Support app no longer relies on
the database.

### DIFF
--- a/app/models/support/requests/anonymous/anonymous_contact.rb
+++ b/app/models/support/requests/anonymous/anonymous_contact.rb
@@ -6,10 +6,8 @@ module Support
     module Anonymous
       class AnonymousContact < ActiveRecord::Base
         before_save :detect_personal_information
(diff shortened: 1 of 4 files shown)


binaryberry:
There is still a url column in the schema - is it still needed?

benilovj:
> There is still a url column in the schema - is it still needed?

good spot... I added b7b6622. I thought we still used `url` somewhere but actually it can be dropped now (tested with Feedback and Support apps).

PR #48, opened 2015-05-14
Endpoint for fetching anonymous feedback by URL

benilovj:
This pull request adds a new endpoint (under `/anonymous-feedback`) for fetching anonymous feedback by path. FeedEx (which used to fetch the feedback directly from the Support Contacts DB) will be the first consumer of this endpoint.

In terms of responsibility, this endpoint overlaps a bit with the `/anonymous-feedback/problem-reports` endpoint; this is temporary, because the `problem-reports` endpoint can be folded into this one as soon as `/anonymous-feedback` can accept a `type` filter.


spec/requests/anonymous_feedback_feed_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+describe "Anonymous feedback feed" do
+  # In order to improve information and services on GOV.UK
+  # As a publisher
+  # I want to view bugs and gripes submitted by GOV.UK users
+
+  context "by path" do
+    let!(:problem_report) do
+      create(:problem_report,
+        what_wrong: "A",
```
binaryberry: Is it normal this problem report is returned even though it is a single word report?

benilovj: yes. The filtering happens in the [`feedback` app](https://github.com/alphagov/feedback/blob/9633c2df78a2b8df1c210faa38d7a5436d503313/app/models/report_a_problem_ticket.rb#L29) and doesn't get as far as `support-api`.

binaryberry: :+1:

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -0,0 +1,25 @@
+class AnonymousFeedbackController < ApplicationController
+  def index
+    unless params[:starting_with_path].present?
```
boffbowsh: `:path_prefix` instead? Or is this prescribed elsewhere

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    query = Support::Requests::Anonymous::AnonymousContact.
+      find_all_starting_with_path(params[:starting_with_path])
+    results = Kaminari.paginate_array(query).
+      page(params[:page]).
+      per(Support::Requests::Anonymous::AnonymousContact::PAGE_SIZE)
+
+    if results.empty?
+      head :not_found
```
boffbowsh: Technically the resource was found, the resource in this case is a query with the given parameters. It'd be better to return `{"results": [], "total_count": 0, ...}` in the case of e.g. a query of `/foobar`. If the user requests a page which we don't have results for, it'd be `{"results": [], "current_page": 200, "pages": 199}`. Either should return a 200 though - a 404 implies the user got the URL wrong.

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    else
+      render json: {
+        results: results,
+        current_page: results.current_page,
+        pages: results.total_pages,
+        page_size: Support::Requests::Anonymous::AnonymousContact::PAGE_SIZE,
+      }
```
boffbowsh: Adding `total_count` in here would be useful

app/models/support/requests/anonymous/anonymous_contact.rb
```diff
@@ -16,11 +16,23 @@ class AnonymousContact < ActiveRecord::Base
...
+        scope :free_of_personal_info, -> {
+          where(personal_information_status: "absent")
+        }
         scope :only_actionable, -> { where(is_actionable: true) }
+
         default_scope {
           only_actionable.
           where.not(path: nil) # this can be removed when path can't be nil
         }
+        scope :find_all_starting_with_path, ->(path) {
```
boffbowsh: Calling this `find_all_…` implies this will return records, but it's a scope. It also hides behaviour, naming it something like `recent_displayable_prefixed_path` might be better, or moving the composition of scopes into the controller would remove ambiguity and enable this to be tested in isolation.

app/models/support/requests/anonymous/long_form_contact.rb
```diff
@@ -7,6 +7,17 @@ class LongFormContact < AnonymousContact
         validates_presence_of :details
         validates :user_specified_url, length: { maximum: 2048 }
         validates :details, length: { maximum: 2 ** 16 }
+
+        def type
+          "long-form-contact"
+        end
+
+        def as_json(options)
+          super(only: [
```
boffbowsh: This is discarding the options passed in to `as_json`.

app/models/support/requests/anonymous/service_feedback.rb
```diff
@@ -8,6 +8,17 @@ class ServiceFeedback < AnonymousContact
         validates :details, length: { maximum: 2 ** 16 }
         validates_inclusion_of :service_satisfaction_rating, in: (1..5).to_a
 
+        def type
+          "service-feedback"
+        end
+
+        def as_json(options)
```
boffbowsh: Ditto above `as_json` commetn

app/controllers/anonymous_feedback_controller.rb
benilovj: agreed, `:path_prefix` is shorter. fixed.

app/controllers/anonymous_feedback_controller.rb
benilovj: agreed. this also makes the result easier to work with. fixed.

app/controllers/anonymous_feedback_controller.rb
benilovj: added.

app/models/support/requests/anonymous/anonymous_contact.rb
benilovj: I've broken up the scope into smaller scopes with individual tests, and chained scopes in the controller. I've also added f4318b1 which gets rid of the default scope, as previously suggested by @evilstreak.

app/models/support/requests/anonymous/long_form_contact.rb
benilovj: fixed.

app/models/support/requests/anonymous/service_feedback.rb
benilovj: fixed.

benilovj:
This PR should be ready for another review.

spec/controllers/anonymous_feedback_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+describe AnonymousFeedbackController do
+  describe "#index" do
+    it "is a bad request with no `path_prefix`" do
+      get :index
+      expect(response).to have_http_status(400)
+    end
+
+    it "is successful with valid `path_prefix`" do
```
boffbowsh: Probably worth a quick additional test for 0 results for this path_prefix, and also another one for the current_page > total_pages case.

spec/models/support/requests/anonymous/anonymous_contact_spec.rb
```diff
@@ -87,6 +83,47 @@ def contact(options = {})
...
+            a = contact(created_at: Time.now - 1.hour)
+            b = contact(created_at: Time.now - 2.hour)
+            c = contact(created_at: Time.now)
+
+            expect(AnonymousContact.most_recent_first).to eq([c, a, b])
+          end
+
+          it "can filter reports with personal information" do
+            a = contact(personal_information_status: "absent")
+            _ = contact(personal_information_status: "suspected")
```
boffbowsh: I'm curious why you assign to `_` here, instead of no assignment at all. Is it just for statement alignment purposes?

boffbowsh:
Other than my suggestion of an additional couple of tests, this looks good.

spec/models/support/requests/anonymous/anonymous_contact_spec.rb
benilovj: I've removed the unnecessary assignment.

spec/controllers/anonymous_feedback_controller_spec.rb
benilovj: I've added the extra tests.

boffbowsh:
:+1: :rocket:

PR #49, opened 2015-05-18
Flatten model modules, remove the Support::Requests::Anonymous namespace

benilovj:
This PR doesn't make any functional modifications to the API, only the following refactorings:

- the `Support::Requests::Anonymous` module has been removed because
it made the directory structure hard to work with and caused lots of
extra nuisance when using the rails console
- require statements featuring models located in `app/models` have been
removed (Rails autoloads those classes automatically)
- two new model files have been created (`TopUrls`, `FeedbackCounts`)
for classes that were housed within the file belonging to another class.
It's best practice for each ruby file to only contain one class.
- the database has been migrated to remove module on anonymous contacts STI type.
This brings the single table inheritance `type` column on the
anonymous_contacts table in line with the flattened model classes.

/cc @binaryberry @mikejustdoit @boffbowsh


app/models/anonymous_contact.rb
```diff
@@ -0,0 +1,57 @@
+require 'field_which_may_contain_personal_information'
```
boffbowsh: Is `lib` not included in autoload?

boffbowsh:
Looks good apart from tiny comment :+1:

app/models/anonymous_contact.rb
benilovj: It seems like if I remove this, then it starts failing (either missing constants when running the specs  or from rake tasks (eg `anonymous_feedback_deduplication:recent`) so I've left it in there for now.

Nonetheless, I fixed a couple of places where I still had `support/requests/anonymous/...` mentioned, which should mean that all rake tasks work now.

benilovj:
This should be ready for another look.

boffbowsh:
Looks good :+1:

PR #50, opened 2015-05-18
Add support for date ranges

binaryberry:
Support-api can now return anonymous contacts for selected date ranges. 
The date ranges are inclusive, can be given in non-chronological order, and the app can handle none, 1 or 2 dates being defined. 
Anonymous contacts also indexed by creation date.
Also moved Timecop.return to spec_helper to ensure it was run after each test.

### DIFF
--- a/app/controllers/anonymous_feedback_controller.rb
+++ b/app/controllers/anonymous_feedback_controller.rb
@@ -8,11 +8,13 @@ class AnonymousFeedbackController < ApplicationController
     from_date = parse_date(params[:from])
     to_date = parse_date(params[:to])
 
+    from_date, to_date = [from_date, to_date].sort if from_date && to_date
+
     results = AnonymousContact.
       only_actionable.
       free_of_personal_info.
       matching_path_prefix(params[:path_prefix]).
-      created_between_days(from_date, to_date).
+      created_between_days(from_date || Time.at(0), to_date || Time.now).
       most_recent_first.
       page(params[:page]).
       per(AnonymousContact::PAGE_SIZE)
--- a/app/controllers/anonymous_feedback_controller.rb
+++ b/app/controllers/anonymous_feedback_controller.rb
@@ -9,6 +9,7 @@ class AnonymousFeedbackController < ApplicationController
       only_actionable.
       free_of_personal_info.
       matching_path_prefix(params[:path_prefix]).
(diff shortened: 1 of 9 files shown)


benilovj:
It's worth updating the author in these commits from the `vagrant` user to your real name...

binaryberry:
Done @benilovj  :-)

spec/models/anonymous_contact_spec.rb
```diff
@@ -112,6 +112,29 @@ def contact(options = {})
...
+      let(:second_date) { Time.new(2014, 10, 31) }
+      let(:third_date){ Time.new(2014, 11, 25) }
+      let(:last_date) { Time.new(2014, 12, 12) }
+
+      before do
+        @first_contact = new_contact(created_at: first_date)
+        @second_contact = new_contact(created_at: second_date)
+        @third_contact = new_contact(created_at: third_date)
+        @newest_contact = new_contact(created_at: last_date + 12.hours)
+        @first_contact.save!
```
benilovj: I think you can do this directly using `contact(attr)` (which builds and saves the feedback), saving one line per contact.

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -5,20 +5,39 @@ def index
...
+    from_date = parse_date(params[:from])
+    to_date = parse_date(params[:to])
+
+    from_date, to_date = [from_date, to_date].sort if from_date && to_date
+
     results = AnonymousContact.
       only_actionable.
       free_of_personal_info.
       matching_path_prefix(params[:path_prefix]).
+      created_between_days(from_date || Time.at(0), to_date || Time.now).
```
benilovj: `Time.now` is going to be in the local timezone, so when we do `date.beginning_of_day` later, this would be the beginning of the day in the local timezone. We ought to work in UTC.

spec/models/anonymous_contact_spec.rb
binaryberry: Thanks! Should be sorted now.

benilovj: :+1:

app/controllers/anonymous_feedback_controller.rb
boffbowsh: Fixed in 060f66b

benilovj:
:+1:

PR #51, opened 2015-05-19
Reinstate necessary require statement

benilovj:
It seems that I was too overzealous in https://github.com/alphagov/support-api/pull/49
and removed a require that's necessary for the worker to run.

This fixes an [errbit issue](https://errbit.preview.alphagov.co.uk/apps/53be40ba0da1154d55001fcb/problems/555a840d65786342cf270000).


boffbowsh:
:+1:

PR #52, opened 2015-05-19
Query the database more efficiently in AnonymousFeedbackController

benilovj:
Before this fix, the page limits weren't being applied to the ActiveRecord
queries because of inappropriate use of Kaminari. This meant that
for paths with lots of feedback, everything was being loaded into
memory before being limited to just one page.

The fix means that the limits are applied directly in the DB query.


boffbowsh:
:shipit:

PR #53, opened 2015-05-20
Endpoint for each organisation with recent feedback counts

benilovj:
This is useful for being able to present FeedEx users with an overview
on which pages have had the most feedback left recently on them.

eg `/anonymous-feedback/organisations/hm-revenue-customs`:
```
{
  slug: "hm-revenue-customs",
  title: "HM Revenue & Customs",
  anonymous_feedback_counts: [
    {
      path: "/government-gateway",
      last_7_days: 0,
      last_30_days: 0,
      last_90_days: 1
    },
   ...
}
```

This is in support of https://tree.taiga.io/project/core-improvement-theme/us/12

@mikejustdoit note that I made a small renaming in the API interface (`results => anonymous_feedback_counts`) which will affect the corresponding Support app code.

/cc @boffbowsh @binaryberry @fofr

### DIFF
--- a/app/controllers/anonymous_feedback/organisations_controller.rb
+++ b/app/controllers/anonymous_feedback/organisations_controller.rb
@@ -9,8 +9,7 @@ module AnonymousFeedback
       organisation = Organisation.find_by(slug: params[:slug])
       anonymous_feedback_counts = ContentItem.
         for_organisation(organisation).
-        summary.
-        sort_by { |r| -1 * r[:last_7_days] }
+        summary("last_7_days")
 
       render json: {
         slug: organisation.slug,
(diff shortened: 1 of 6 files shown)


boffbowsh:
I've updated this with a small performance improvement. Hope you don't mind, it was one of those "I think this might be quicker?" moments, and working out whether I was correct or not got me most of the way to implementing it. I also ended up adding ordering support.

Everything else looks good to me :+1:

benilovj:
@binaryberry you mind taking a look at this today, since a few of us added code to this...

boffbowsh:
Verbal :+1: from @binaryberry

PR #54, opened 2015-05-21
Add rake task to import Organisations from Whitehall

boffbowsh:
Steals an example from the needs_api. Will require a new set of
credentials creating for whitehall API and adding to the deployment
scripts.

Also added new acronym and govuk_status fields to organisations.

/cc @benilovj @mikejustdoit

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -7,6 +7,7 @@ class Organisation < ActiveRecord::Base
   validates :slug, presence: true
   validates :web_url, presence: true
   validates :title, presence: true
+  validates :govuk_status, presence: true
 
   def self.for_path(path)
     orgs_data = SupportApi::enhanced_content_api.organisations_for(path) || []
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -7,7 +7,6 @@ class Organisation < ActiveRecord::Base
   validates :slug, presence: true
   validates :web_url, presence: true
   validates :title, presence: true
-  validates :govuk_status, presence: true
 
   def self.for_path(path)
     orgs_data = SupportApi::enhanced_content_api.organisations_for(path) || []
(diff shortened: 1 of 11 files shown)


config/initializers/gds_api.rb
```diff
@@ -0,0 +1,6 @@
+# These credentials are used to import organisations
+# using OrganisationImporter
+
+ORGANISATIONS_API_CREDENTIALS = {
+  bearer_token: ENV["ORGANISATIONS_API_CREDENTIALS"] || "change me when you deploy"
```
jamiecobbett: The organisations API is publicly accessible: https://www.gov.uk/api/organisations

jamiecobbett:
Could this not pull on the fly from content-register, the way that [policy-publisher does](https://github.com/alphagov/policy-publisher/blob/master/app/models/policy.rb#L79)?

boffbowsh:
re: content register, I'm not sure. @benilovj?

boffbowsh:
Content Register doesn't appear to include the slug, abbreviation or status, so we'll need to use Whitehall for now.

benilovj:
As @jamiecobbett, we probably ought to incorporate `content_id` at this point and use that as the primary key for the sync, rather than `slug`, unless there's a very good reason not to. That way we'll have continuity when slugs change.

boffbowsh:
Ok, will do that.

boffbowsh:
@jamiecobbett @benilovj I've made that change, we now key off the content_id. I've added a possibly-contentious db migration which looks up against content register. It doesn't do this if there are no organisations in the DB, to make things easier for devs starting afresh.

benilovj:
:+1: what's there LGTM

One possible last thing to add is an entry [in the schedule](https://github.com/alphagov/support-api/blob/master/config/schedule.rb) to run this daily (?).

boffbowsh:
Added, @benilovj

PR #57, opened 2015-05-28
Add Export to CSV option for Feedback reports

boffbowsh:
Takes a date range, a path prefix and an email address to notify as parameters. It then enqueues a Sidekiq job which generates the CSV and dumps it in `/data/uploads/feedback_explorer`. This is a mounted NFS path and will be available to both frontend and backend boxes. The link in the email will eventually make the support app ask nginx to send the correct CSV to the user using the `X-Accel-Redirect` header.

Work to come to support this on the front end:
- [x] Support app work to add an Export to CSV button, and a controller endpoint to serve the generated CSV
- [x] Puppet work to add the `/data/uploads/feedback_explorer` location to nginx.
- [x] Api adapter work to add the new endpoints
- [x] puppet work to ensure the directory exists and is writable

/cc @benilovj @evilstreak @bradleywright

### DIFF
--- a/app/mailers/export_notification.rb
+++ b/app/mailers/export_notification.rb
@@ -1,0 +1,10 @@
+class ExportNotification < ActionMailer::Base
+  default from: "inside-government@digital.cabinet-office.gov.uk"
+
+  layout false
+
+  def notification_email(email, url)
+    @url = url
+    mail(to: email, subject: "[Feedback Explorer] Your feedback export is ready")
+  end
+end
--- a/app/mailers/export_notification.rb
+++ b/app/mailers/export_notification.rb
@@ -5,6 +5,6 @@ class ExportNotification < ActionMailer::Base
 
   def notification_email(email, url)
     @url = url
-    mail(to: email, subject: "[Feedback Explorer] Your feedback export is ready")
+    mail(to: email, subject: "[GOV.UK Feedback Explorer] Your feedback export is ready")
   end
 end
--- a/app/models/anonymous_contact.rb
+++ b/app/models/anonymous_contact.rb
@@ -18,6 +18,7 @@ class AnonymousContact < ActiveRecord::Base
   }
   scope :only_actionable, -> { where(is_actionable: true) }
   scope :most_recent_first, -> { order("created_at DESC") }
+  scope :most_recent_last, -> { order("created_at ASC") }
   scope :matching_path_prefix, ->(path) { where("path LIKE ?", path + "%") }
   scope :created_between_days, -> (first_date, last_date) { where(created_at: first_date..last_date.at_end_of_day) }
 
--- a/app/models/anonymous_contact.rb
+++ b/app/models/anonymous_contact.rb
@@ -21,6 +21,17 @@ class AnonymousContact < ActiveRecord::Base
   scope :matching_path_prefix, ->(path) { where("path LIKE ?", path + "%") }
   scope :created_between_days, -> (first_date, last_date) { where(created_at: first_date..last_date.at_end_of_day) }
 
+  scope :for_query_parameters, ->(options={}) do
+    path_prefix = options[:path_prefix] || "/"
+    from = options[:from] || Date.new(1970)
+    to = options[:to] || Date.today
+
+    only_actionable.
+      free_of_personal_info.
+      matching_path_prefix(path_prefix).
+      created_between_days(from, to)
+  end
+
   PAGE_SIZE = 50
   paginates_per PAGE_SIZE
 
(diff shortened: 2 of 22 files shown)


app/models/anonymous_contact.rb
```diff
@@ -18,9 +18,21 @@ class AnonymousContact < ActiveRecord::Base
   }
   scope :only_actionable, -> { where(is_actionable: true) }
   scope :most_recent_first, -> { order("created_at DESC") }
+  scope :most_recent_last, -> { order("created_at ASC") }
   scope :matching_path_prefix, ->(path) { where("path LIKE ?", path + "%") }
   scope :created_between_days, -> (first_date, last_date) { where(created_at: first_date..last_date.at_end_of_day) }
 
+  scope :for_query_parameters, ->(options={}) do
+    path_prefix = options[:path_prefix] || "/"
```
benilovj: I have a slight preference for changing the `matching_path_prefix` scope to only work if `path_prefix` is provided, rather than having a default for `path_prefix`...

app/mailers/export_notification.rb
```diff
@@ -0,0 +1,10 @@
+class ExportNotification < ActionMailer::Base
+  default from: "inside-government@digital.cabinet-office.gov.uk"
+
+  layout false
+
+  def notification_email(email, url)
+    @url = url
+    mail(to: email, subject: "[Feedback Explorer] Your feedback export is ready")
```
benilovj: Might be worth mentioning GOV.UK in the subject.

app/models/anonymous_contact.rb
boffbowsh: Makes sense.

app/models/feedback_export_request.rb
```diff
@@ -0,0 +1,41 @@
...
+    csv = CSV.new(io)
+    csv << FeedbackCsvRowPresenter::HEADER_ROW
+    results.find_each do |row|
+      csv << FeedbackCsvRowPresenter.new(row).to_a
+    end
+    io
+  end
+
+  def url
+    Plek.find('support') + "/anonymous-feedback/export-requests/#{id}"
```
benilovj: because of an annoying inconsistency between the `support` and `support-api` apps, this isn't quite right - the namespace in the `support` app is `/anonymous_feedback` (with an underscore). Is there any way to reduce coupling between the support and support-api apps? It may be possible to use gds-api-adapters to generate the URL to the export requests within `support` (that way if the URL structure changes in `support`, we wouldn't need to make changes here) but that may be going a bit over the top...

db/migrate/20150522151256_create_feedback_export_requests.rb
```diff
@@ -0,0 +1,13 @@
+class CreateFeedbackExportRequests < ActiveRecord::Migration
+  def change
+    create_table :feedback_export_requests do |t|
+      t.string :notification_email
```
benilovj: worth enforcing not-null constraints on this table?

benilovj:
:+1: apart from the minor comments.

app/models/feedback_export_request.rb
boffbowsh: I'm not sure how best to decouple this without the support-api calling the support app. Even putting it in gds-api-adapters means it's coupled in that way. Maybe make the support app send the email instead of the api?

bradleywright: @boffbowsh how would it do that, since this request is asynchronous?

boffbowsh: Another worker in the support app? But yeah, also not ideal at all. I don't see an easy way of decoupling this.

boffbowsh: Are there precedents for API apps knowing the locations of things in the frontend apps?

benilovj:
One way to break the dependency:

- the support app polls the support-api every few mins
- if there's a completed export, it emails the user and acknowledges the job

boffbowsh:
Artificially slow user experience :(. Technically we're not introducing a dependency on the support app itself, just that whatever is on that URL does the correct thing. We could always add redirects later if needed.

benilovj:
ok let's leave it as is for now, :+1: ready to merge when all the prerequisites are in place.

boffbowsh:
Supporting work in https://github.com/alphagov/gds-api-adapters/pull/310 and https://github.gds/gds/puppet/pull/2844

benilovj:
LGTM :shipit: when the puppet work has been merged.

app/views/export_notification/notification_email.text.erb
```diff
@@ -0,0 +1,5 @@
+Hi,
+
+The CSV you requested from the Feedback Explorer is available here:
```
benilovj: Worth getting a content designer to review this email?

PR #60, opened 2015-06-05
Allow filtering feedback by Organisation

boffbowsh:
Taiga: https://tree.taiga.io/project/core-improvement-theme/us/52

Problem Reports are categorised by Organisation, so they should be filterable by Organisation too. Passing an `organisation_slug` parameter to the API will only return Problem Reports associated with that Organisation. Service Feedback and Long Form Contacts aren't currently associated with an org, so they will never be returned.

This includes some refactoring to support multiple future types of filter for CSVs, by storing the filter parameters as a serialised hash in the database.


bishboria:
:+1:

PR #61, opened 2015-06-08
Refactor organisation-related endpoints

boffbowsh:
Move the Organisation index into a separate controller as it’s
unrelated to anonymous feedback
Add an Organisation show action to retrieve the title etc for a single
organisation
Add a spec for the org summary action

Backwards-compatible for the support app as the `/anonymous-feedback/organisations` endpoint still works, but should be removed.

### DIFF
--- a/app/controllers/anonymous_feedback/organisations_controller.rb
+++ b/app/controllers/anonymous_feedback/organisations_controller.rb
@@ -1,10 +1,5 @@
 module AnonymousFeedback
   class OrganisationsController < ApplicationController
-    def index
-      organisations = Organisation.order(:slug)
-      render json: organisations
-    end
-
     def show
       if %w(path last_7_days last_30_days last_90_days).include? params[:ordering]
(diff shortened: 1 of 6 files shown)


app/controllers/anonymous_feedback/organisations_controller.rb
```diff
@@ -12,6 +7,12 @@ def show
         ordering = "last_7_days"
       end
       organisation = Organisation.find_by(slug: params[:slug])
+
+      if organisation.nil?
+        render nothing: true, status: :not_found
+        return
+      end
```
mikejustdoit: This (the part of the spec that covers it) could go in a separate commit if they're as unrelated to the other changes as it seems

boffbowsh: I did it all as one as I wrote the spec from scratch - it didn't exist before.

mikejustdoit:
Looks good

PR #63, opened 2015-06-11
Support MySQL and PostgreSQL

boffbowsh:
We need to move this app to PostgreSQL, mainly due to MySQL's query planner not being able to use multiple indexes but also in order to support future ML/sentiment analysis work, which is easier with some of the Postgres native data types.

Rough migration plan, please also review this:

1. Deploy this. Nothing should change.
2. Stop Sidekiq workers, so no new data is inserted
3. Copy data, perhaps using https://github.com/maxlapshin/mysql2postgres? Fallback to exporting CSVs and importing using `COPY FROM`
4. Reboot app with `SUPPORT_API_DB_TYPE=postgresql`, test.
5. Restart Sidekiq workers with `SUPPORT_API_DB_TYPE=postgresql`, check new data is inserted correctly.
6. Make `SUPPORT_API_DB_TYPE` changes permanent in `puppet` / `alphagov_deployment`.

The bulk of the work here is courtesy of @benilovj, I added the ability to use both DBs interchangeably, namely the backported schema dumper fix and the "logic" in database.yml.

/cc @jennyd.


boffbowsh:
Forgot to mention: needs the deployment repo updating to make the same changes to the secret database.yml for Postgres support. MySQL will continue to work if this is merged but no deployment changes are made.

jennyd:
We used [this script](https://github.com/rgarner/mysql-postgresql-converter/tree/drop_before_create) to convert Transition's database dump when we migrated to Postgres last year; we were converting a full dump with schema, though, not just the data.

jennyd:
There are some differences in behaviour between MySQL and Postgres which [we had to deal with in Transition](https://github.com/alphagov/transition/pull/436) and you may need to deal with here. In particular:

- In practice MySQL was returning results in a consistent order when no ordering was specified, but that wasn't the case with Postgres. We found that we had been depending on the order being consistent in various places (both code and tests) and had to change all of those to be explicit about ordering.
- We found that we had been relying on MySQL's case insensitivity in queries to find hostnames in Bouncer, so needed to [be explicit about that](https://github.com/alphagov/bouncer/commit/3b93f9f5d21f603b63c0d084c719d9e83a8d2da7) too.

jennyd:
The deployment plan sounds reasonable, but I'd recommend testing the app on Postgres in a production-like environment for a while before deploying. We found that Postgres performance is really dependent on the environment, and things that worked well for Transition in a dev environment on a single machine with SSD needed some changes to perform acceptably in production.

It's also important to test your method of converting the dump, both to make sure it will load correctly and to see how long it will take in production. Increasing some memory-related settings may speed up the loading - `maintenance_work_mem` was relevant for us, but may be less useful for you if you're only loading the data, not creating indexes from the dump.

`wal_keep_segments` is already set to 128 for all our postgres servers so you shouldn't have any problems with replication not being able to keep up while the dump is loading.

boffbowsh:
Updated for case sensitivity, @jennyd

app/models/anonymous_contact.rb
```diff
@@ -22,7 +22,11 @@ class AnonymousContact < ActiveRecord::Base
   scope :only_actionable, -> { where(is_actionable: true) }
   scope :most_recent_first, -> { order("created_at DESC") }
   scope :most_recent_last, -> { order("created_at ASC") }
-  scope :matching_path_prefix, ->(path) { where("anonymous_contacts.path LIKE ?", path + "%") if path }
+  scope :matching_path_prefix, ->(path) do
+    column = SupportApi.postgresql? ? "lower(anonymous_contacts.path)" : "anonymous_contacts.path"
```
benilovj: Do we actually need to do this? I understand that it's now possible to use case-insensitive paths but I'm not convinced that anyone was expecting or relying on case-insensitivity in paths...

boffbowsh: mindblown

You make a good point. Looking through the data, there's many upper case characters in the paths, except a few levels down where users won't be filtering anyway.

Shall I get rid of this then?

benilovj: > Shall I get rid of this then?

:+1: that's the simpler solution.

boffbowsh:
I've now removed the case insensitivity, and added a rake task to do the data synchronisation. Please review, @jennyd and @benilovj

boffbowsh:
The sync takes ~30 minutes on dev. Next step is to test on preview.

benilovj:
:+1:

PR #69, opened 2015-07-07
Remove MySQL integration

benilovj:
The Support API has been running fine on PostgreSQL in production for some time,
and the MySQL integration can now be dropped.

The PostgreSQL integration uses an SQL-based schema dump, so the ruby version
of the schema can be dropped (it was only being used by MySQL).

This PR also clears out code used during the DB migration.

/cc @boffbowsh


mattbostock:
\o/

PR #72, opened 2015-09-22
Owning organisation(s) for each problem report now updated each time one comes in

benilovj:
Before this change, the owning organisation for each piece of content was only determined when the content was first seen. This meant that for content where the owning org was updated, the changes were never picked up.

https://trello.com/c/WK0EqQUl/100-content-in-feedex-incorrectly-appearing-for-dvsa-filter-medium

/cc @jamiecobbett @boffbowsh @binaryberry


jamiecobbett:
:top:

PR #73, opened 2015-10-13
Bump gds_zendesk gem, no longer set the date fields on Zendesk tickets

benilovj:
The custom date fields on Zendesk tickets are no longer set
for other Zendesk tickets, and in this app these fields were
never populated anyway, so we can safely not send them to Zendesk.

I've tested this by:
- pushing this branch to Preview
- raising a problem report
- making sure that the ticket lands in Zendesk successfully

/cc @binaryberry @boffbowsh


benlovell:
LGTM. :ship:

jamiecobbett:
Thanks @benilovj

PR #74, opened 2015-11-02
Make pagination more performant

boffbowsh:
For queries that will return a lot of results, fetching the initial result count is very expensive (20s for over 3 million results).

Luckily, Rails can do a `COUNT` with `LIMIT` in PostgreSQL. If you specify a limit as part of a scope then call `.count`, it will generate SQL like this:

```sql
SELECT COUNT(count_column) FROM 
  (SELECT 1 AS count_column FROM "anonymous_contacts" 
   WHERE path LIKE '/done/vehicle-tax%' 
   LIMIT 10000) subquery_for_count;
```

This returns in around 20ms on a dev VM for a query that would return 3 million rows if unlimited. It provides 10000 responses over 200 pages for the user, with the last page rendering in about 100ms.

If the user requires all responses, they can still export them all as a CSV. Currently only 28 paths have more than 10000 responses, all of which are Done pages which need to be refactored at some point to display only those with comments.

This should be coupled with frontend changes, providing copy informing the user to download a CSV if there's more than 10000 responses.

/cc @tijmenb @benilovj

### DIFF
--- a/app/controllers/anonymous_feedback_controller.rb
+++ b/app/controllers/anonymous_feedback_controller.rb
@@ -15,6 +15,7 @@ class AnonymousFeedbackController < ApplicationController
 
     json[:from_date] = dates[0] if dates[0]
     json[:to_date] = dates[1] if dates[1]
+    json[:results_limited] = results_limited if results_limited
 
     render json: json
   end
@@ -38,6 +39,10 @@ class AnonymousFeedbackController < ApplicationController
 
   def total_count
     @total_count ||= scope.limit(AnonymousContact::MAX_PAGES * AnonymousContact::PAGE_SIZE).count
+  end
+
+  def results_limited
+    total_count >= AnonymousContact::PAGE_SIZE * AnonymousContact::MAX_PAGES
   end
 
   def pages
--- a/app/controllers/anonymous_feedback_controller.rb
+++ b/app/controllers/anonymous_feedback_controller.rb
@@ -5,31 +5,57 @@ class AnonymousFeedbackController < ApplicationController
       return
     end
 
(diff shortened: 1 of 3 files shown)


benilovj:
Seems legit so far, although as you already mentioned, will need careful refactoring since the implementation is somewhat involved.

boffbowsh:
Updated, @benilovj, @benlovell @jamiecobbett

benilovj:
This looks sensible, although I'd personally have a couple of unit tests to make sure that the calculations of `pages` and `offset` are free from off-by-1 errors.

boffbowsh:
Added some tests to cover the above, and a flag that indicates that the results were truncated.

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -5,31 +5,62 @@ def index
...
+    AnonymousContact.
+      for_query_parameters(path_prefix: params[:path_prefix],
+                           organisation_slug: params[:organisation_slug],
+                           from: dates[0],
+                           to: dates[1]).
+      most_recent_first
+  end
+
+  def dates
+    [parse_date(params[:from]), parse_date(params[:to])].tap do |dates|
```
benlovell: This method could be simplified to `[parse_date(params[:from]), parse_date(params[:to])].sort!` as the postfix modifier seems redundant.

benilovj:
:+1: looks good. Feel free to merge (when the FeedEx frontend has been updated?).

boffbowsh:
We're probably going to just deploy this as-is as it fixes a broken thing, then add some UI stuff later.

benlovell:
:ship: nice.

PR #76, opened 2015-12-03
Improvements around organisation lookup

benilovj:
This is a collection of refactorings and code improvements to make this area of the code easier to understand and to work with.

The highlights:
- better naming of classes and methods
- extract code specifically relating to talking to Content API and Content Store
- fetch the `content_id` when looking up orgs

/cc @jamiecobbett

PR #77, opened 2015-12-03
Use content_ids to reconcile orgs between the saved DB and the Content Store/API

benilovj:
This PR:
- makes `content_id` mandatory for all orgs
- cleans up the org without a `content_id`
- uses `content_id`s when dynamically creating orgs that aren't imported yet in the nightly import. This allows the reconciliation which happens during import to work correctly.

Open TODOs:

- [x] this relies on #76 and should be rebased after that's merged

~~f64eff7 should probably move the code out of the migration into a more generic data hygiene class that's unit tested~~

/cc @jamiecobbett

### DIFF
--- a/db/migrate/20151202212408_clean_up_org_without_content_id.rb
+++ b/db/migrate/20151202212408_clean_up_org_without_content_id.rb
@@ -1,0 +1,22 @@
+class CleanUpOrgWithoutContentId < ActiveRecord::Migration
+  def change
+    orgs_without_a_content_id = Organisation.where(content_id: nil).count
+    raise "Unexpected number (#{orgs_without_a_content_id}) of organisations without a content_id" if orgs_without_a_content_id > 1
+
+    old_org = Organisation.find_by(slug: 'schools-commissioner')
+    new_org = Organisation.find_by(slug: "schools-commissioners-group")
+    if old_org && new_org
+      old_org.content_items.each do |item|
+        item.organisations.delete(old_org)
+        item.organisations << new_org
+      end
+
+      old_org.reload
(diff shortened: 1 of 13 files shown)


boffbowsh:
Rebased now that #76 is merged

db/migrate/20151202212408_clean_up_org_without_content_id.rb
```diff
@@ -0,0 +1,22 @@
+class CleanUpOrgWithoutContentId < ActiveRecord::Migration
+  def change
+    orgs_without_a_content_id = Organisation.where(content_id: nil).count
+    raise "Unexpected number (#{orgs_without_a_content_id}) of organisations without a content_id" if orgs_without_a_content_id > 1
+
+    old_org = Organisation.find_by(content_id: nil)
```
jamiecobbett: This feels a little dangerous. Is there an alternative that would reveal the intent better? Could we look for it by the old slug?

jamiecobbett:
:+1: 

> f64eff78c5f99b63a43d56fa7c5e2e02f0cf835b should probably move the code out of the migration into a more generic data hygiene class that's unit tested

Are we likely to do this again?

benilovj:
> Are we likely to do this again?

That's a fair point. Because we're using content_ids now, orgs should hopefully be renamed seamlessly and this process shouldn't be necessary in future. I'll leave the code as is, and it can be extracted later if necessary.

db/migrate/20151202212408_clean_up_org_without_content_id.rb
benilovj: Agreed. Fixed and force-pushed.

benilovj:
This is ready for another review.

PR #82, opened 2016-05-20
Aggregate service feedbacks and remove from Anonymous contacts table

binaryberry:
Service feedbacks are ratings that are requested of users at the end of a transaction - at the moment they represent 16 million rows out of the 17 million rows of anonymous contacts, and therefore slow down queries considerably, to the point they sometimes timeout. 

This removes service feedback ratings from the anonymous contacts table unless they have attached comments, and saves them to a separate table, the ArchivedServiceFeedback table. It also aggregates the count and adds rows to the feedex table so that users can see how many ratings of 1, 2, 3, 4, and 5 they received each day for each path.

Tested in integration and works with [this support PR](https://github.com/alphagov/support/pull/277).

Before:
screen shot 2016-05-27 at 17 18 00

After:
screen shot 2016-06-06 at 11 52 10

After this, we will need to run the aggregation task on all existing rows, run Vacuum Analyze in the database in production, and double check the aggregation task gets run daily.

[Trello](https://trello.com/c/zhqhIKOs/377-reduce-duplication-of-quantitative-results-in-feedex-medium)


db/structure.sql
```diff
@@ -271,41 +319,6 @@ CREATE INDEX index_anonymous_contacts_on_created_at ON anonymous_contacts USING
 
 
 --
--- Name: index_anonymous_contacts_on_created_at_and_path; Type: INDEX; Schema: public; Owner: -; Tablespace:
```
binaryberry: @boffbowsh a bit puzzled as to why running the migration removed some indexes... is that normal?

boffbowsh: This means that your database schema that it is created this structure from is missing these indexes. Perhaps they didn't get restored correctly from the integration backup during replication. Best to remove these lines from your commit.

lib/tasks/service_feedback_aggregation.rake
```diff
@@ -0,0 +1,40 @@
...
+
+    aggregate(date_range)
+  end
+
+  def aggregate(date_range)
+    start_time = Time.now
+    p "Processing feedback from #{date_range.first} to #{date_range.last}"
+    date_range.each do |date|
+      aggregator = ServiceFeedbackAggregator.new
+      aggregator.run(date)
```
boffbowsh: Might be good to print these dates to the console as they're run, and timestamped so whoever is running it can get an idea of how quickly it's running.

lib/tasks/service_feedback_aggregation.rake
```diff
@@ -0,0 +1,40 @@
...
+    end_date = Date.parse(STDIN.gets.chomp)
+
+    date_range = start_date..end_date
+
+    aggregate(date_range)
+  end
+
+  def aggregate(date_range)
+    start_time = Time.now
+    p "Processing feedback from #{date_range.first} to #{date_range.last}"
```
boffbowsh: Should use `puts` rather than `p`.

lib/tasks/service_feedback_aggregation.rake
```diff
@@ -0,0 +1,40 @@
...
+    date_range.each do |date|
+      aggregator = ServiceFeedbackAggregator.new
+      aggregator.run(date)
+    end
+    end_time = Time.now
+    p "Service Feedback has been aggregated and extracted to a different table"
+    p "duration = #{end_time - start_time}"
+  end
```
boffbowsh: Might be good to run a Postgres [`VACUUM ANALYZE`](http://www.postgresql.org/docs/9.3/static/sql-vacuum.html) statement on `anonymous_contacts` at the end of the run. Once everything has been rolled up, we can schedule a `VACUUM FULL` for out of hours to fully reclaim all the space this operation is saving.

README.md
```diff
@@ -20,3 +20,7 @@ to populate this in your development environment:
 
 1. Ensure whitehall is running: `bowl whitehall`
 2. Run the rake task `rake api_sync:import_organisations`
+
+## Data
+
+When [replicating data](https://github.gds/gds/development#8-import-production-data), by default the anonymous feedback is not imported. You will have to import it by going into development and running `./sync-postgresql.sh -r -s -d backups/YYYY-MM-DD postgresql-primary-1.backend.integration`, where the date corresponds to the date of your latest backup.
```
fofr: Good instructions! :100:

db/structure.sql
binaryberry: I did now! The diff is still a bit polluted by what I think is my editor's automated removal of spaces at the end of line. Hopefully it's no biggie but let me know if that was an issue.

boffbowsh: Yeah, can you remove those changes from the diff please? Else next time the schema is regenerated it'll all just change again.

lib/tasks/service_feedback_aggregation.rake
binaryberry: Done now in bb0844e. For each date, outputting the date + duration of the task and removed total duration as it seemed superfluous.

lib/tasks/service_feedback_aggregation.rake
binaryberry: Thanks! addressed now.

db/structure.sql
binaryberry: Done now :-)

app/models/aggregated_service_feedback.rb
```diff
@@ -0,0 +1,18 @@
+class AggregatedServiceFeedback < AnonymousContact
+
+  def type
+    "aggregated-service-feedback"
+  end
+
+  def as_json(options = {})
+    attributes_to_serialise = [
```
Davidslv: you can use `%i`  

```
> %i(a b)
=> [:a, :b]
```

Not a blocker.

lib/service_feedback_aggregator.rb
```diff
@@ -0,0 +1,71 @@
+class ServiceFeedbackAggregator
+
+  def run(date)
+    @date = date
```
Davidslv: This looks weird here.
You still have to initialize `ServiceFeedbackAggregator` so why not give it the date there?
Otherwise this feels like a class method.

binaryberry: Good point @Davidslv, and I had initially written it like that. The reason I opted for this way was that in the rake task that uses this, we are going to process large amounts of service feedback, and if we only have to instantiate the ServiceFeedbackAggregator once instead of once per date, I think we'll save a bit of time? Not dead set on this though.

lib/service_feedback_aggregator.rb
```diff
@@ -0,0 +1,71 @@
...
+      true,
+      'absent',
+      DATE_TRUNC('day',created_at),
+      DATE_TRUNC('day',created_at),
+      service_satisfaction_rating,
+      count(service_satisfaction_rating)
+      FROM anonymous_contacts
+      WHERE type = 'ServiceFeedback'
+      AND created_at >= '#{@date}'
+      AND created_at < '(#{@date + 1.day})'
```
boffbowsh: What are the parentheses for here?

lib/tasks/service_feedback_aggregation.rake
```diff
@@ -0,0 +1,40 @@
...
+      aggregate(date_range)
+      Rails.logger.info "Daily service feedback aggregation has finished"
+    end
+  end
+
+  desc "Extract service feedback of specified date to a different table, and add aggregated results"
+  task :date => :environment do
+    require 'service_feedback_aggregator'
+
+    puts "Please enter the date you would like aggregation to start from, in the format YYYY-MM-DD:"
```
boffbowsh: Can we parameterise these for the rake task please, then we can run the script from within Jenkins.

lib/service_feedback_aggregator.rb
boffbowsh: I think @Davidslv is right, the class should be initialised once per date, then have `run` called on it. This ensures that any state doesn't leak between runs. Instantiation of a PORO like this is relatively inexpensive.

spec/models/service_feedback_aggregator_spec.rb
```diff
@@ -0,0 +1,100 @@
+require 'rails_helper'
+require 'service_feedback_aggregator'
+
+describe ServiceFeedbackAggregator do
+  let(:date)  { Time.new(2013,2,10,10)  }
```
boffbowsh: You've declared `date` helpers but there's a few places your'e still using the literal date in the code below

spec/models/service_feedback_aggregator_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        end
+      end
+    end
+  end
+
+  context "archiving service feedbacks" do
+    let(:date) { Time.new(2013,2,11) }
+
+    it "copies service feedbacks to the archived service feedback table" do
+      expect(ArchivedServiceFeedback.count).to eq 0
```
boffbowsh: You can use the expect change matcher for these: http://www.relishapp.com/rspec/rspec-expectations/v/3-4/docs/built-in-matchers/change-matcher

spec/models/service_feedback_aggregator_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  let!(:service_feedback_aggregator) { ServiceFeedbackAggregator.new }
+
+  context "aggregating service feedbacks" do
+
+    before { service_feedback_aggregator.run(Time.new(2013,2,10)) }
+
+    context "when there are two service feedbacks" do
+      context "with the same rating" do
+        it "there is one aggregated service feedback with a rating of 1" do
```
boffbowsh: This doesn't read correctly. The example should read as english, including the `it`.
```ruby
it "creates an aggregated service feedback with a rating of 1" do
```

spec/models/service_feedback_aggregator_spec.rb
```diff
@@ -0,0 +1,100 @@
+require 'rails_helper'
+require 'service_feedback_aggregator'
+
+describe ServiceFeedbackAggregator do
+  let(:date)  { Time.new(2013,2,10,10)  }
+  let(:slug)  { "done/register-to-vote" }
+  let(:rating){ 1                       }
```
boffbowsh: It's ok to not have braces align. Personally I think the whitespace here looks overly forced.

spec/models/service_feedback_aggregator_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  let!(:service_feedback_aggregator) { ServiceFeedbackAggregator.new }
+
+  context "aggregating service feedbacks" do
+
+    before { service_feedback_aggregator.run(Time.new(2013,2,10)) }
+
+    context "when there are two service feedbacks" do
+      context "with the same rating" do
+        it "there is one aggregated service feedback with a rating of 1" do
+          expect(AggregatedServiceFeedback.all.map(&:service_satisfaction_rating)).to eq [1]
```
boffbowsh: You can use `pluck` here instead of `map`

app/models/aggregated_service_feedback.rb
binaryberry: Better indeed. I would prefer to keep it this way to conserve consistency with the other models ([long_form_contact](https://github.com/alphagov/support-api/blob/master/app/models/long_form_contact.rb#L10), [organisation](https://github.com/alphagov/support-api/blob/master/app/models/organisation.rb#L11), [problem report](https://github.com/alphagov/support-api/blob/master/app/models/problem_report.rb#L21) and [service feedback](https://github.com/alphagov/support-api/blob/master/app/models/service_feedback.rb#L10)). The styleguide says that "Consistency with this style guide is important. Consistency within a project is more important."

lib/service_feedback_aggregator.rb
binaryberry: To test your eyes: you passed! Amended in relevant commits.

lib/service_feedback_aggregator.rb
binaryberry: Gotcha! Thanks for the explanation. Amended in 4850364.

lib/tasks/service_feedback_aggregation.rake
binaryberry: Sure, done in 4850364.

spec/models/service_feedback_aggregator_spec.rb
binaryberry: Good point, done in 4850364.

spec/models/service_feedback_aggregator_spec.rb
binaryberry: Nice! Addressed in 4850364.

spec/models/service_feedback_aggregator_spec.rb
binaryberry: Indeed! Amended now in 4850364.

spec/models/service_feedback_aggregator_spec.rb
binaryberry: oki dok! Amended in 4850364.

spec/models/service_feedback_aggregator_spec.rb
binaryberry: Done, thanks! 4850364

binaryberry:
@boffbowsh I believe I've addressed all comments, and I also added 60b2287 to avoid duplicate service feedback aggregates

lib/service_feedback_aggregator.rb
```diff
@@ -0,0 +1,88 @@
+class ServiceFeedbackAggregator
+
+  def initialize(date)
+    @date = date
+  end
+
+  def run
+    return reason_for_not_running if reason_for_not_running
```
boffbowsh: This doesn't get exposed anywhere

binaryberry: Thanks! Added in 7ac6928.

lib/service_feedback_aggregator.rb
boffbowsh: This one is still here

lib/service_feedback_aggregator.rb
```diff
@@ -0,0 +1,93 @@
+class ServiceFeedbackAggregator
+
+  attr_reader :reason_for_not_running
+
+  def initialize(date)
+    @date = date
+  end
+
+  def run
+    return @reason_for_not_running if check_reason_for_not_running
```
gpeng: This use the accessor `reason_for_not_running` rather than the variable `@reason_for_not_running`

binaryberry: Thanks! Addressed now.

binaryberry:
@boffbowsh : I addressed the remaining parenthesis in the SQL query, and updated the daily rake task to only include yesterday's date, not today's. Should be good to go now, hopefully!

PR #86, opened 2016-07-05
Remove headers argument

mgrassotti:
Remove a redundant headers argument from the sidekiq workers.

This was needed because the sidekiq server middleware was missing in tests. 
This has now been added to the new version of the govuk_sidekiq gem and therefore this argument is not needed anymore.


boffbowsh:
👍  merge on Thursday

PR #87, opened 2016-07-13
Add aggregated feedback details to CSV exports

binaryberry:
CSV exports showed an empty cell for aggregated feedback lines, instead of the "Rating of 1: 12" type details field that is displayed in the view. This fixes it and ensures CSV exports include this details field.

This also improves two existing tests:
1. When testing different ratings, we shouldn't expect results in a specific order. This
test was failing intermittently.
2. The test for different dates only had different times, not a different day. This
fixes it.

This work was carried out pairing with @mgrassotti, and tested on integration by deploying this branch, and then exporting a CSV for a /done/ page as well as a regular page.

[Trello](https://trello.com/c/TVlBgbol/425-feedex-aggregation-not-adding-up-correctly)

### DIFF
--- a/app/presenters/feedback_csv_row_presenter.rb
+++ b/app/presenters/feedback_csv_row_presenter.rb
@@ -19,7 +19,7 @@ class FeedbackCsvRowPresenter
       row.created_at.strftime("%F %T"),
       row.type == "service-feedback" ? row.slug : row.path,
       details_text,
-      row.type == "service-feedback" ? row.service_satisfaction_rating.to_s : "",
+      service_satisfaction_rating,
       parsed_user_agent.family,
       parsed_user_agent.version.to_s,
       parsed_user_agent.os.family,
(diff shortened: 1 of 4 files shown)


app/presenters/feedback_csv_row_presenter.rb
```diff
@@ -41,6 +41,17 @@ def details_text
       row.details
     when "long-form-contact"
       row.details
+    when "aggregated-service-feedback"
+      "Rating of #{row.service_satisfaction_rating}: #{row.details}"
+    end
+  end
+
+  def service_satisfaction_rating
+    if row.type == "service-feedback" ||
```
boffbowsh:
```ruby
if row.type =~ /(aggregated-)?service-feedback/
```

app/presenters/feedback_csv_row_presenter.rb
```diff
@@ -41,6 +41,16 @@ def details_text
       row.details
     when "long-form-contact"
       row.details
+    when "aggregated-service-feedback"
+      "Rating of #{row.service_satisfaction_rating}: #{row.details}"
```
boffbowsh: The service satisfaction rating is already included, so having this as text makes it less parseable in Excel. It'd be better just to include the number of records with this rating, so just `row.details`.

app/presenters/feedback_csv_row_presenter.rb
binaryberry: 👍

app/presenters/feedback_csv_row_presenter.rb
binaryberry: Good point, but I'm not sure how understandable it will be to have just a number in a column called "feedback" and whether users will understand what the number represents?

PR #88, opened 2016-08-02
Bring dependencies up to date

kalleth:
Let's be good citizens and as we touch the application bring its dependencies up to date.

This includes a unicorn upgrade so I think that has implications on deployment.


boffbowsh:
For deployment, it'll just need to be deployed with a hard restart. I don't think there's a huge risk in doing this, we'll just get a handful of 502s during the deployment

elliotcm:
Be careful with that GDS API adapters upgrade, it technically includes 6 breaking changesets, though in reality they're unlikely to have touched the code used by this app.

Worth reviewing the changelog if you haven't already.

kalleth:
@elliotcm Yup -- as far as I can see in the app we only use `GdsApi::Support`, `GdsApi::PerformancePlatform::DataIn` and `GdsApi::Organisations`, none of which have changed since ~2014.

kalleth:
:shipit:

boffbowsh:
For future reference, you should only self-merge if someone gave a "merge when green", or a 👍 for Whitehall

kalleth:
@boffbowsh Assumed that because both you and Elliot had cast your eyes over it you were fine with it as I'd addressed the only comment. However, I'll wait in future.

PR #89, opened 2016-08-04
Allow export of Global statistics

kalleth:
[Trello card](https://trello.com/c/ptXseJEJ/13-export-global-anonymous-feedback-csv)

This generates and emails (as an attachment) a CSV file containing two
fields; `date` and `report_count`, for all anonymous feedback submitted
to GOV.UK.

The aim here is to provide a mechanism to export this data so that
statistical analysis can be performed. At the moment this analysis
happens on data from zendesk; we'd like to avoid sending all types of
anonymous feedback to zendesk, so we need to have another way to provide
the data.

### DIFF
--- a/app/controllers/anonymous_feedback/global_export_requests_controller.rb
+++ b/app/controllers/anonymous_feedback/global_export_requests_controller.rb
@@ -1,0 +1,27 @@
+class AnonymousFeedback::GlobalExportRequestsController < ApplicationController
+  def create
+    export_request = GlobalExportRequest.new(global_export_request_params)
+    if export_request.valid?
+      GenerateGlobalExportCsvWorker.perform_async(global_export_request_params)
+      render nothing: true, status: 202
+    else
+      render json: { "errors" => export_request.errors.to_a }, status: 422
+    end
+  end
+
+private
+  def global_export_request_params
+    permitted_params = [
+      :from_date,
+      :to_date,
+      :notification_email
+    ]
+
+    clean_params = params.require(:global_export_request).permit(*permitted_params)
+    {
+      from_date: parse_date(clean_params[:from_date]),
+      to_date: parse_date(clean_params[:to_date]),
+      notification_email: clean_params[:notification_email],
+    }
+  end
+end
--- a/app/models/global_export_csv_generator.rb
+++ b/app/models/global_export_csv_generator.rb
@@ -1,0 +1,33 @@
+class GlobalExportCsvGenerator
+  def initialize(from_date, to_date)
+    @from_date = from_date
+    @to_date = to_date
+  end
+
+  def call
+    return filename, generate_csv
+  end
+
+private
+  attr_reader :from_date, :to_date
+
+  def results
+    ProblemReport.
+      created_between_days(from_date, to_date).
+      select("date(created_at) as created_at_date, COUNT(id) as report_count").
+      group("created_at_date").
+      order("created_at_date").
+      limit(10_000)
+  end
+
+  def generate_csv
+    CSV.generate do |csv|
+      csv << ['date', 'report_count']
+      results.each { |r| csv << [r[:created_at_date].strftime('%Y-%m-%d'), r[:report_count]] }
+    end
+  end
+
+  def filename
+    "feedex_#{from_date.iso8601}_#{to_date.iso8601}.csv"
+  end
+end
--- a/app/controllers/anonymous_feedback/export_requests_controller.rb
+++ b/app/controllers/anonymous_feedback/export_requests_controller.rb
@@ -30,8 +30,8 @@ class AnonymousFeedback::ExportRequestsController < ApplicationController
       clean_params = params.require(:export_request).permit(*permitted_params)
       {
         filters: {
-          from: parse_date(clean_params[:from]),
-          to: parse_date(clean_params[:to]),
+          from: parse_date(clean_params[:from_date]),
+          to: parse_date(clean_params[:to_date]),
           path_prefix: clean_params[:path_prefix],
           organisation_slug: clean_params[:organisation]
         },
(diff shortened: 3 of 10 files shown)


app/models/global_export_csv_generator.rb
```diff
@@ -0,0 +1,43 @@
...
+  attr_reader :from, :to
+
+  def results
+    # ProblemReport.
+    #   created_between_days(from, to).
+    #   select("date(created_at), COUNT(id) as report_count").
+    #   group("date(created_at)").
+    #   order("date(created_at)").
+    #   limit(10_000)
+    sql = <<-ESQL
```
kalleth: Ideally there would be a way to use the former (commented out, activerecord explanation) of this query.

However, ActiveRecord when using it's query interface directly won't allow you to use clauses that return calculated fields, as far as I can see, as they're not attributes on the given model.

Suggestions welcome.

elliotcm: AR will return any calculated field so long as you alias it, it just won't show it in the string representation of the object.  For example:

```
irb(main):010:0> reports = ProblemReport.created_between_days(Date.parse("2016-08-01"), Date.parse("2016-08-04")).select("date(created_at) as created_at_date, COUNT(id) as report_count").group("date(created_at)").order("date(created_at)").limit(10_000)
=> #<ActiveRecord::Relation [#<ProblemReport id: nil>, #<ProblemReport id: nil>, #<ProblemReport id: nil>, #<ProblemReport id: nil>]>
irb(main):011:0> reports.last.created_at_date
=> Thu, 04 Aug 2016
irb(main):012:0> reports.last.report_count
=> 447
```

elliotcm: You can also use those aliases in the group and order clauses if you wish.

app/controllers/anonymous_feedback/global_export_requests_controller.rb
```diff
@@ -0,0 +1,22 @@
+class AnonymousFeedback::GlobalExportRequestsController < ApplicationController
+  def create
+    export_request = GlobalExportRequest.new(global_export_request_params)
+    if export_request.valid?
+      GenerateGlobalExportCsvWorker.perform_async(global_export_request_params)
+      render nothing: true, status: 202
```
elliotcm: Should we return some sort of basic JSON success message in case a human is triggering these directly?

app/controllers/anonymous_feedback/global_export_requests_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    else
+      render json: { "errors" => export_request.errors.to_a }, status: 422
+    end
+  end
+
+private
+  def global_export_request_params
+    permitted_params = %i(from to notification_email)
```
elliotcm: Can you stack these vertically so it's easier to see changes in git in the future

app/controllers/anonymous_feedback/global_export_requests_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    clean_params = params.require(:global_export_request).permit(*permitted_params)
+    {
+      from: parse_date(clean_params[:from]),
+      to: parse_date(clean_params[:to]),
```
elliotcm: Should we be clearer in the API that these are dates, i.e. any provided time data will be dropped?  e.g. `from_date`, `to_date`

app/models/global_export_csv_generator.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def generate_csv
+    CSV.generate do |csv|
+      csv << ['date', 'report_count']
+      results.each { |r| csv << [r[:date].strftime('%Y-%m-%d'), r[:report_count]] }
+    end
+  end
+
+  def filename
+    "feedex_#{from.to_date.iso8601}_#{to.to_date.iso8601}.csv"
```
elliotcm: Is it worth adding the current time into the filename too?  Can you request a date range twice and get different contents? e.g. requesting yesterday and today while today is still happening

elliotcm:
Looking good, a few questions / comments.

app/controllers/anonymous_feedback/global_export_requests_controller.rb
kalleth: I'd forgotten in my comment in the previous review that the time component was stripped at this level, so they will always be dates. Now I feel sheepish.

app/controllers/anonymous_feedback/global_export_requests_controller.rb
kalleth: I've decided against this based on the rest of the application:

```
$ grep -ir "render nothing" app/controllers
app/controllers/anonymous_feedback/export_requests_controller.rb:      render nothing: true, status: 202
app/controllers/anonymous_feedback/global_export_requests_controller.rb:      render nothing: true, status: 202
app/controllers/anonymous_feedback/long_form_contact_controller.rb:        render nothing: true, status: 202
app/controllers/anonymous_feedback/organisations_controller.rb:        render nothing: true, status: :not_found
app/controllers/anonymous_feedback/problem_reports_controller.rb:        render nothing: true, status: 202
app/controllers/anonymous_feedback/service_feedback_controller.rb:        render nothing: true, status: 202
app/controllers/organisations_controller.rb:      render nothing: true, status: :not_found
```

Whilst i might decide later on to do this, for now (as it's my first PR on the app) I'm going to hold off modifying these on all the controllers.

app/models/global_export_csv_generator.rb
kalleth: It is possible, yes -- I could do, but I'm not sure it's worth it as the primary use case of this feature is for historical data for the previous week after the week has completed.

The feedback explorer export doesn't do this.

elliotcm:
👍 Good to merge when ready.

PR #90, opened 2016-08-09
Group organisation report summaries by path

kalleth:
# [Trello card](https://trello.com/c/7kwsLiHh/22-duplicates-in-feedex)

## Before
screenshot 2016-08-09 16 33 13

## After
screenshot 2016-08-09 16 33 25

## WTF?

In the database for the `support-api`, it's possible for there to be multiple `ContentItem`'s for a given `path`. Case in point:

    irb(main):005:0> ContentItem.where(path: '/calculate-your-holiday-entitlement').count
    => 255

When a problem report is raised, a representation of the content item relating to that problem report is persisted to the database.

The process of accessing this content item is pretty convoluted, and reading the code it would *appear* to be trying to update the existing content item in the database if one already exists, but it appears that this isn't happening.

Unpicking this code requires a significant refactor and some reasonable understanding of the GOV.UK content API/content stores/publishing API, which I don't have yet, but in the meantime let's fix the user facing bug.

I imagine we'll revisit this as part of revisiting the data model of problem reports.

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -23,7 +23,7 @@ class ContentItem < ActiveRecord::Base
       select("#{last_30_days} AS last_30_days").
       select("#{last_90_days} AS last_90_days").
       where("anonymous_contacts.created_at > ?", midnight_last_night - 90.days).
-      group("content_items.id").
+      group("content_items.path").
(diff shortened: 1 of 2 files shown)


elliotcm:
Add a test and you're good to go.

elliotcm:
👍

benilovj:
Huzzah!

PR #92, opened 2016-08-16
Refactor the ServiceFeedbackAggregator and use it to send data to the performance platform

kalleth:
# [Trello card](https://trello.com/b/zdXw6qbx/survey-support-team-doing)

This is aiming to fix a bug where the performance platform received less numbers of service feedback than were reported in the feedex UI.

Better off reviewing commit by commit as my reasoning for each change is explained in the commit messages. Although this is a "net positive" in lines, I think the resultant code is far simpler to understand when reading it.

These need completing before merge:

- [x] Ensure that changing `config/schedule.rb` will change existing cron jobs on a deployment
- [x] Ensure there's no problem with sending data to the performance platform an hour later than usual
- [x] Ensure service aggregation will *always* be complete after an hour.


spec/models/service_feedback_spec.rb
```diff
@@ -14,11 +14,14 @@
...
+    result_attrs = ServiceFeedback.aggregates_by_rating.inject({}) do |m, r|
+      m[r[:service_satisfaction_rating]] = r[:cnt]
+      m
+    end
+
+    expect(result_attrs).to eq(
       1 => 1,
       2 => 1,
-      3 => 0,
-      4 => 0,
```
elliotcm: Is this behaviour still tested at the ServiceFeedbackAggregatedMetrics level?

lib/service_feedback_aggregated_metrics.rb
```diff
@@ -12,13 +12,20 @@ def to_h
...
-    results_array = by_rating.map {|rating, count| ["rating_#{rating}", count] }.flatten +
-                    [ "comments", filter_by_day_and_slug.with_comments_count ] +
-                    [ "total", by_rating.values.inject(:+)]
-
-    Hash[*results_array]
+    {
+      "comments" => filter_by_day_and_slug.with_comments_count,
+      "total" => ratings.values.inject(:+) || 0,
+    }.tap do |result|
+      (1..5).each { |i| result["rating_#{i}"] = (ratings[i] || 0) }
```
NeilvB: I don't feel strongly about it but I wonder if we're gaining all that much from using a loop here rather than simple mappings of each rating key. This means we can move to eg 10 star ratings without having to change this code but on the other hand, using simple mappings means we can remove the `tap` and there's a small increase in readability of how the hash is constructed IMO.

spec/lib/service_feedback_aggregator_spec.rb
```diff
@@ -69,6 +69,21 @@
...
+          create(:duplicate_service_feedback,
+                 service_satisfaction_rating: 1,
+                 slug: "register-to-vote",
+                 created_at: date,
+                )
+        end
+
+        it "doesn't include it in the count" do
+          aggregator.run
+          expect(AggregatedServiceFeedback.pluck(:details)).to eq ["2"]
```
elliotcm: This could be a simple count instead of a pluck.

It's also not clear by this point why it's 2 and not 1, could you give it its own context with a more bespoke setup?

elliotcm:
Looks generally good.

lib/service_feedback_aggregated_metrics.rb
```diff
@@ -12,13 +12,20 @@ def to_h
...
+    end
+  end
+
+  def ratings
+    @results ||= Hash[
+      filter_by_day_and_slug.
```
NeilvB: Do you reckon we could move this into the model? I wonder if `actionable_feedback_by_day_and_slug(day, slug)` or similar might be a more readable name.

spec/models/service_feedback_spec.rb
kalleth: This entire class is removed in a subsequent commit (the final one) and replaced by using the AggregatedServiceFeedback totals which are themselves tested.

spec/lib/service_feedback_aggregator_spec.rb
kalleth: That's a good point; I took some of the internals of the test from the original spec file.

Within the `AggregatedServiceFeedback` context, the contents of "details" is the sum of the number of `ServiceFeedback` entries that had this `service_satisfaction_rating`.

So a count would return `1` here because there's a single entry. This is possibly a sign that the description of the test needs to be changed, so I'll look at doing that.

lib/service_feedback_aggregated_metrics.rb
```diff
@@ -12,13 +12,20 @@ def to_h
...
+        aggregates_by_rating.
+        reduce({}) { |m, r| m[r[:service_satisfaction_rating]] = r[:cnt]; m }
```
NeilvB: Might just be me not fully understanding - but here we're using `m` for `memo` right - but what does `memo` mean?

lib/service_feedback_aggregator.rb
```diff
@@ -17,16 +17,18 @@ def run
 
 private
 
-  def check_reason_for_not_running
+  def raise_if_should_not_run
```
NeilvB: Good idea 👍

lib/service_feedback_aggregator.rb
```diff
@@ -54,6 +56,7 @@ def create_aggregate
       count(service_satisfaction_rating)
       FROM anonymous_contacts
       WHERE type = 'ServiceFeedback'
+      AND is_actionable = true
```
binaryberry: 👍

lib/service_feedback_aggregated_metrics.rb
kalleth: I feel exactly the same way -- which is why in a subsequent commit I removed it :)

lib/service_feedback_aggregated_metrics.rb
kalleth: Removed entirely in a subsequent commit :)

lib/service_feedback_aggregated_metrics.rb
kalleth: Memo represents the collection that starts off as what's passed to 'reduce' (`{}`) and is passed to and added to in each iteration of the loop.

I agree, though, this wasn't clear. It's been removed, but I should have written this clearer from the get-go.

kalleth:
For ~32k records:

```
deploy@integration-backend-1:/var/apps/support-api$ time govuk_setenv support-api bundle exec rake service_feedback_aggregation:daily
Processing feedback from 2016-08-16 to 2016-08-16
Aggregation complete for 2016-08-16. Duration: 0.83094646
Service Feedback has been aggregated and extracted to a different table

real    0m3.487s
user    0m2.096s
sys     0m0.766s
```

I suspect we'll be OK here -- if it takes 3.5 seconds on integration for 32k records, giving it an hour seems safe.

kalleth:
Following the above I've verified that running the rake task directly pushes data to the performance platform. Good to go.

kalleth:
Also verified that the original bug was fixed by comparing https://www.preview.performance.service.gov.uk/data/vehicle-tax/customer-satisfaction?flatten=true&period=week&start_at=2016-08-16T00%3A00%3A00Z&end_at=2016-08-16T23%3A59%3A59Z&collect=rating_1%3Asum&collect=rating_2%3Asum&collect=rating_3%3Asum&collect=rating_4%3Asum&collect=rating_5%3Asum&collect=total%3Asum to https://support.integration.publishing.service.gov.uk/anonymous_feedback?page=48&path=%2Fdone%2Fvehicle-tax and making sure that the totals reported for the 16th of August (yesterday) match.

elliotcm:
Good to merge!

PR #93, opened 2016-08-19
Correctly parse dates for export requests

kalleth:
[Trello card](https://trello.com/c/lkPUHwQD/31-csv-export-on-feedback-includes-all-feedback-rather-than-just-the-dates-requested)

When performing an export request, we should use 'from' and 'to', not 'from_date' and 'to_date', as these are not present in the incoming hash.

This means that export requests have *never* respected dates.


NeilvB:
Looks good 👍

PR #94, opened 2016-09-01
Review Problem Reports endpoints

NeilvB:
Part of work on [Mark Anonymous Feedback as Spam](https://trello.com/c/g1I2TXF6/14-mark-anonymous-feedback-as-spam-download-filtered-report-xxl).

Adds two new endpoints to the `support-api` app:
  - An endpoint for marking specified problem reports as "spam"
  - An endpoint for fetching pages of problem report lists with various filters

The endpoint for fetching reports replaces an existing endpoint (routed to `ProblemReports#index`). This was replaced as the Kibana logs indicate that it is not in use:

screen shot 2016-09-01 at 17 13 07

Do not merge until other dependent branches are reviewed.


app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -58,6 +66,24 @@ def problem_report_params
       )
     end
 
+    def mark_supplied_reports_as_reviewed_and_spam
+      reviewed_problem_reports_hash.each_pair do |problem_report_id, marked_as_spam|
+        problem_report = ProblemReport.find_by_id(problem_report_id)
```
kalleth: Instead of iterating through the array and finding each one individually here, we could find by the array of values:

```ruby
def mark_supplied_reports_as_reviewed_and_spam
  ProblemReport.find(reviewed_problem_reports_hash.keys).each do |report|
    review_attrs = { reviewed: true, marked_as_spam: reviewed_problem_reports_hash[report.id] }
    unless problem_report.update(review_attrs) #etc
  end
  true
rescue ActiveRecord::RecordNotFound # find([array]) will raise if any value not found
  return false
end
```

Providing an ID which isn't present in the DB should be exceptional, hence rescuing this isn't a problem. I think.

db/migrate/20160826105129_add_reviewed_to_anonymous_contacts.rb
```diff
@@ -0,0 +1,5 @@
+class AddReviewedToAnonymousContacts < ActiveRecord::Migration
+  def change
+    add_column :anonymous_contacts, :reviewed, :boolean, null: false, default: false
```
kalleth: I'd probably combine these into a single migration. Doesn't matter, functionally, though.

kalleth:
Looks good :+1: -- couple of comments. We're not quite at the stage where we should refactor the controller to remove logic from it into tested helper classes, but we're getting close. Perhaps the next PR (providing the index!) is a potential target for this.

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -22,21 +21,26 @@ def totals
...
-              end
-      @results = query.where(created_at: @interval)
-
-      if @results.empty?
-        head 204
+      offset = (current_page - 1) * AnonymousContact::PAGE_SIZE
+      limit = AnonymousContact::PAGE_SIZE
+
+      from_date = parse_date(params[:from_date]) || Date.new(1970)
+      to_date = parse_date(params[:to_date]) || Date.today
```
kalleth: Perhaps Time.now? `Date.today` returns midnight, iirc (`00:00:00`) We might want to make sure we don't miss reports from today.

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -22,21 +21,26 @@ def totals
     end
 
     def index
-      query = if @selected_organisation
```
kalleth: We're confident this wasn't used in feedex elsewhere, right?

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -22,21 +21,26 @@ def totals
...
+      offset = (current_page - 1) * AnonymousContact::PAGE_SIZE
+      limit = AnonymousContact::PAGE_SIZE
+
+      from_date = parse_date(params[:from_date]) || Date.new(1970)
+      to_date = parse_date(params[:to_date]) || Date.today
+
+      exclude_reviewed = params[:exclude_reviewed] || false
+
+      if exclude_reviewed
+        results = ProblemReport.all.created_between_days(from_date, to_date).most_recent_first.where(reviewed: false).offset(offset).limit(limit).most_recent_first
```
kalleth: Yeah, this is definitely starting to feel like a class needs extracting here. I feel like we're going to be able to test this more easily if it's in a class who's sole point is to accept a set of params and return a set of results, which the controller presents as a JSON structure.

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -22,21 +21,26 @@ def totals
...
-        end
+        results = ProblemReport.all.created_between_days(from_date, to_date).most_recent_first.offset(offset).limit(limit)
       end
+
+      json = {
+        results: results,
+        current_page: current_page
+      }
+
+      render json: json
```
kalleth: nit: `render json: {results: results, current_page: current_page}` would be fine here

config/routes.rb
```diff
@@ -40,8 +40,7 @@
         constraints: { date: /\d{4}-\d{2}-\d{2}/ },
         to: 'problem_reports#totals'
 
-    get '/problem-reports/:period',
```
kalleth: I'm still nervous about removing this, but i guess if we're sure it's not used.... Did we look through Kibana? I think we did...

NeilvB: I know what you mean. I did a search through the logs and there were 0 calls listed to this endpoint: 
screen shot 2016-09-01 at 17 13 07

app/controllers/anonymous_feedback/problem_reports_controller.rb
NeilvB: Nice. Thanks.

app/controllers/anonymous_feedback/problem_reports_controller.rb
NeilvB: Yes good shout! Thanks

NeilvB: Actually, on further examination, the called scope calls `.at_end_of_day` on this so that's covered:

app/controllers/anonymous_feedback/problem_reports_controller.rb
NeilvB: Added a service type class for the controller to call.

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -50,14 +37,46 @@ def create
       end
     end
 
+    def mark_reviewed_for_spam
+      if mark_supplied_reports_as_reviewed_and_spam
+        render nothing: true, status: 200
```
elliotcm: Probably worth rendering some simple JSON here for the sake of humans reading it, and for consistency so consumers can parse the response with a JSON parser and not get errors.

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -50,14 +37,46 @@ def create
...
+      else
+        render nothing: true, status: 404
+      end
+    end
+
   private
+    def problem_report_index_params
+      params.permit(:from_date, :to_date, :exclude_reviewed, :page)
```
elliotcm: Do we `permit` or do we `require` these?

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -50,14 +37,46 @@ def create
...
       params.require(:problem_report).permit(
         :path, :referrer, :javascript_enabled, :user_agent, :what_doing,
         :what_wrong, :source, :page_owner
       )
     end
 
+    def mark_supplied_reports_as_reviewed_and_spam
+      begin
+        ProblemReport.find(reviewed_problem_reports_hash.keys).each do |problem_report|
+          review_attrs = { reviewed: true, marked_as_spam: reviewed_problem_reports_hash[problem_report.id.to_s] }
```
elliotcm: This hash lookup is a bit opaque, what's the intent?

app/controllers/anonymous_feedback/problem_reports_controller.rb
NeilvB: Good question. I feel that `permit` is what suits this endpoint. The only filter on the results that is necessary is the `page`, but we default to the first page anyway so I don't think it's necessary to require it here. I think it's going to be easier to work on the calling app ('support') without making these required.

app/controllers/anonymous_feedback/problem_reports_controller.rb
NeilvB: True - I've added a new variable to store this value from the lookup - it's a boolean indicating whether the report was marked as spam.

NeilvB:
Thanks for the reviews @kalleth and @elliotcm! I've addressed the main questions that you raised.

elliotcm:
Good to go.

kalleth:
:shipit:

PR #96, opened 2016-09-15
Send page improvements to Zendesk

thehenster:
# Why

The service manual wants a way to allow users to give positive feedback, as well as negative.

# What

This receives feedback requests from the feedback frontend on behalf of the 'Improve this page' widget and forwards the feedback to Zendesk. https://trello.com/c/HN1L3EfK/286-update-footer-feedback-component

There isn't a requirement to browse the feedback anywhere apart from on Zendesk so we avoid the "Anonymous Feedback" flow in this app.

The 'Improve this page' feature is an experiment on the Service Manual. Traffic is expected to be very low. Therefore performance hasn't been taken seriously. Notably we are communicating with Zendesk synchronously. This will make the widget less snappy but has the advantage of being able to report to the user if it was successful.

### DIFF
--- a/spec/models/page_improvement_spec.rb
+++ b/spec/models/page_improvement_spec.rb
@@ -1,11 +1,11 @@
 require 'rails_helper'
 
 describe PageImprovement, 'validations' do
-  it 'validates presence of path' do
+  it 'validates presence of url' do
     page_improvement = described_class.new({})
     page_improvement.valid?
 
-    expect(page_improvement.errors.messages).to include(path: include("can't be blank"))
+    expect(page_improvement.errors.messages).to include(url: include("can't be blank"))
   end
 
   it 'validates presence of description' do
@@ -19,7 +19,7 @@ end
 describe PageImprovement, '#zendesk_ticket_attributes' do
   it "generates a hash of attributes to create a Zendesk ticket" do
     page_improvement = described_class.new({
-      path: '/service-manual/test',
+      url: 'https://gov.uk/service-manual/test',
       description: 'I love this page.',
       name: 'John',
       email: 'john@example.com',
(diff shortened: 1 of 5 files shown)


spec/models/page_improvement_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      description: 'I love this page.',
+      name: 'John',
+      email: 'john@example.com',
+      user_agent: 'Safari',
+    })
+
+    expect(page_improvement.zendesk_ticket_attributes).to eq({
+      'subject' => '/service-manual/test',
+      'comment' => {
+        'body' => "[Details]\nI love this page.\n\n[Name]\nJohn\n\n[Email]\njohn@example.com\n\n[Path]\n/service-manual/test\n\n[User agent]\nSafari\n"
```
36degrees: For the sake of readability and clarity on what we expect, is it worth making this and the body in the following block a heredoc?

spec/requests/page_improvements_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    expect(response.code).to eq('201')
+    expect(response_hash).to include('status' => 'success')
+  end
+
+  it "sends the feedback to Zendesk" do
+    zendesk_request = expect_zendesk_to_receive_ticket(
+      "subject" => "/service-manual/agile",
+      "comment" => {
+        "body" => "[Details]\nI have a problem\n\n[Name]\nJohn\n\n[Email]\njohn@example.com\n\n[Path]\n/service-manual/agile\n\n[User agent]\nSafari\n"
```
36degrees: Same thing re heredoc for clarity?

thehenster: Thanks. I've replaced both of these with heredocs.

NeilvB:
Looks good 👍

PR #98, opened 2016-10-05
Refactor content item enrichment

kalleth:
The content item enrichment is ending up with multiple content items with the same path in the database. This makes no sense.

I've not been able to work out why -- the code seemed pretty straightforward.

Instead, I've refactored how content item lookups happen so that they return data structures only, which are then interpreted into the database records in a single place (the enrichment worker). This means that there can't be any side effects outside this class that result in duplicates being created.

This should reduce the frequency of it happening again while I bash my head against a brick wall trying to figure out why.

### DIFF
--- a/app/workers/content_item_enrichment_worker.rb
+++ b/app/workers/content_item_enrichment_worker.rb
@@ -12,9 +12,15 @@ private
   def fetch_content_item(path)
     looked_up_item = SupportApi.content_item_lookup.lookup(path)
     if content_item = ContentItem.find_by(path: path)
-      content_item.tap { |item| item.organisations = looked_up_item.organisations } # refresh the orgs
+      content_item.tap { |item| item.organisations = build_orgs(looked_up_item.organisations) } # refresh the orgs
     else
-      looked_up_item
+      ContentItem.new(path: looked_up_item.path, organisations: build_orgs(looked_up_item.organisations))
+    end
+  end
+
+  def build_orgs(org_hashes)
+    org_hashes.map do |org_data|
+      Organisation.create_with(org_data).find_or_create_by(content_id: org_data[:content_id])
     end
   end
 end
--- a/lib/content_api_lookup.rb
+++ b/lib/content_api_lookup.rb
@@ -5,7 +5,11 @@ class ContentAPILookup
 
   def lookup(path)
     response = api_response(path)
-    ContentItem.new(path: URI(response["web_url"]).path, organisations: organisations_from(response)) if response
+
+    LookedUpContentItem.new(
+      path: URI(response['web_url']).path,
+      organisations: organisations_from(response),
+    ) if response.present?
   end
 
   def organisations_from(response)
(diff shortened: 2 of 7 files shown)


NeilvB:
This looks OK to me but it would be good to have someone with more experience of the content related apps and how they're managed in other apps to take a look too.

lib/content_api_lookup.rb
```diff
@@ -19,20 +23,19 @@ def organisations_from(response)
...
       }
-      Organisation.create_with(org_info).find_or_create_by(content_id: org_info[:content_id])
-    }
+    end
   end
 
   def api_response(path)
     slug = path[1..-1] # the content API expects "vat-rates" instead of "/vat-rates"
-    return nil if slug.nil? || slug.empty?
+    return nil if slug.nil? || slug.empty? # this could be done further up
```
boffbowsh: Can we either remove this comment or do what it says please? This is more a TODO than a warning comment.

lib/content_store_lookup.rb
```diff
@@ -6,43 +6,45 @@ def initialize(content_store)
...
+      response_links["organisations"],
+      response_links["lead_organisations"],
+      response_links["supporting_organisations"],
+    ].compact.flatten.uniq
   end
 
-  def find_or_create_org(org_info)
-    Organisation.create_with(org_info).find_or_create_by(content_id: org_info[:content_id])
+  def organisation?(response)
+    response["format"].gsub("placeholder_", "") == "organisation"
```
boffbowsh: You can just use `response["document_type"] == "organisation"` now. `format` is deprecated. Example: https://www.gov.uk/api/content/government/organisations/land-registry

lib/looked_up_content_item.rb
```diff
@@ -0,0 +1,41 @@
...
+  # or the Content API, but in case we can't, determine a sensible default
+  def guess_organisations_for(path)
+    org_slug = if path =~ %r{^/government/world/organisations}
+                 case path
+                 when /dfid/ then "department-for-international-development"
+                 when /uk-trade-investment/ then "uk-trade-investment"
+                 else "foreign-commonwealth-office"
+                 end
+               elsif path =~ %r{^/government/organisations/hm-revenue-customs/contact}
+                 # TODO: remove this once Contacts Admin populates `links#organisations` in the Content Store
```
boffbowsh: This is now true: https://www.gov.uk/api/content/government/organisations/hm-revenue-customs/contact/agent-dedicated-line-debt-management

app/workers/content_item_enrichment_worker.rb
```diff
@@ -12,9 +12,15 @@ def perform(problem_report_id)
   def fetch_content_item(path)
     looked_up_item = SupportApi.content_item_lookup.lookup(path)
     if content_item = ContentItem.find_by(path: path)
-      content_item.tap { |item| item.organisations = looked_up_item.organisations } # refresh the orgs
+      content_item.tap { |item| item.organisations = build_orgs(looked_up_item.organisations) } # refresh the orgs
     else
-      looked_up_item
+      ContentItem.new(path: looked_up_item.path, organisations: build_orgs(looked_up_item.organisations))
```
NeilvB: My feeling is that is mainly a question of coding style as we're currently instantiating the model here and then saving it just the once in the calling method above. Otherwise we'd be creating and then saving the model again in quick succession. I suppose that it also means that if more validations are added to the model that may be violated here these would all be surfaced from the same place following the one call to `save`... How strongly do you feel about having this saved here too?

NeilvB:
Thanks for the careful review and additional links @boffbowsh  I've addressed most of the questions in the code.

app/workers/content_item_enrichment_worker.rb
boffbowsh: Ah sorry, I didn't realise that. Fine then 👍

kalleth:
👍 woooooo!

PR #100, opened 2016-10-26
Handle 410 (gone/archived) content item responses

NeilvB:


h-lame:
Adding specs too!  A+++ would merge again.

PR #101, opened 2016-10-26
Backfill historic aggregated service feedback

NeilvB:
[Trello card](https://trello.com/c/r09CtbDd/48-user-satisfaction-score-for-a-page-does-not-match-the-feedex-data)

Users have noticed a discrepancy for Service Feedback totals between data presented in the performance platform app and that presented in “Feedex”.

Performance platform has historically displayed data that excludes feedback items that was detected as duplicate. This results in lower service feedback counts than that displayed in Feedex, which does include duplicate items in its counts.

A previous PR tackled this problem for all new service feedback items from 18th August 2016 (https://github.com/alphagov/support-api/pull/92).

This PR aims to give us the ability to fix this discrepancy for historical items prior to this date.

This script is intended to be run once to overwrite historic `AggregatedServiceFeedback` within a certain historical date range within Feedex with data returned from the performance platform.

The result of running this script is that it should fix all historic data discrepancies between Feedex and performance platform (for those `AggregatedServiceFeedback` items for which there is an equivalent, valid data set in the performance platform).


lib/backfill_historic_aggregated_service_feedback.rb
```diff
@@ -0,0 +1,86 @@
...
+        all_service_feedback_hash
+      end
+    rescue GdsApi::HTTPNotFound
+      @logger.warn("No endpoint found in performance platform for #{transaction_slug}")
+      nil
+    end
+  end
+
+  def performance_platform_data_out
+    @data_out ||= GdsApi::PerformancePlatform::DataOut.new("http://www.performance.service.gov.uk")
```
h-lame: I don't know enough about the performance platform to know this, but can' we use `Plek` here?  By hardcoding the url this means no matter what environment we run this on, we talk to the production performance platform.

Maybe that's ok though?

lib/backfill_historic_aggregated_service_feedback.rb
```diff
@@ -0,0 +1,86 @@
...
+    if all_aggregated_feedback_paths.include?(path_for_transaction_slug(transaction_slug))
+      true
+    else
+      @logger.warn("No aggregated feedback found for path #{path_for_transaction_slug(transaction_slug)}")
+      false
+    end
+  end
+
+  def all_aggregated_feedback_paths
+    @@paths ||= AggregatedServiceFeedback.select(:path).distinct.pluck(:path)
```
h-lame: Do you really mean to use a class variable here?  An instance variable is probably fine.

spec/lib/backfill_historic_aggregated_service_feedback_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+      expect(end_date_aggregated_service_feedback_rating_4.reload.details).to eq '9'
+    end
+
+    it 'deletes aggregates in the case where the performance platform returns a count of zero' do
+      expect(AggregatedServiceFeedback.find_by_id(end_date_aggregated_service_feedback_rating_5.id)).to be_falsy
+    end
+
+    it 'does not change the aggregated feedback outside of the specified dates' do
+      expect(earlier_aggregated_service_feedback.changed?).to eq false
+      expect(later_aggregated_service_feedback.changed?).to eq false
```
h-lame: I don't think asking these instances if they've `changed?` is going to do what we want.  `changed?` acts on the specific instance, so will only say yes if that instance had attributes changed, it doesn't reload the values from the db and compare the instance and db values.  The `BackfillHistoricAggregatedServiceFeedback` class loads `AggregatedServiceFeedback` instances directly from the db, and I can't see that we've stubbed these load calls, so we'll have to check that they haven't changed in the same way that we check the ones that have changed (reloading from the db and expecting the values to remain the same).

This comment goes for the next example too.

NeilvB:
Thanks for the review @h-lame. I addressed the two main issues and force-pushed my changes.

spec/lib/backfill_historic_aggregated_service_feedback_spec.rb
NeilvB: Thanks for pointing this out

PR #104, opened 2016-12-19
Fix Airbrake initializer

steventux:


sihugh:
Hurrah!

config/initializers/airbrake.rb
```diff
@@ -1,5 +1,12 @@
...
-  config.development_environments << "production"
+  if ENV.has_key?("ERRBIT_API_KEY")
+    config.api_key          = ENV["ERRBIT_API_KEY"]
+    config.host             = "errbit.#{ENV['GOVUK_APP_DOMAIN']}"
+    config.secure           = true
+    config.environment_name = ENV['ERRBIT_ENVIRONMENT_NAME']
+  else
+    # Adding "production" to the development environments causes Airbrake not
+    # to attempt to send notifications.
+    config.development_environments << "production"
```
tijmenb: We don't have this in the other airbrake initialisers. Do you know what it's for?

chao-xian: @tijmenb you added your review comment at the same minute @sihugh merged. Is this ok to deploy? Cc @steventux

steventux: This was originally the 'development' config for Airbrake (ie. for running locally) in real deployments this config would have been overwritten. I struggled to understand what it does but _I think_ it's to suppress Airbrake when running RAILS_ENV=production locally.
Maybe it should just go.

PR #106, opened 2017-01-23
Move to new ci

mgrassotti:
Trello: https://trello.com/c/iGeDwcSz/582-move-support-api-to-new-ci-infrastructure-1

The old CI infrastructure is in the process of being wound down so we need to move our apps to the new one.


Jenkinsfile
```diff
@@ -0,0 +1,100 @@
...
+
+    stage("git merge") {
+      govuk.mergeMasterBranch()
+    }
+
+    stage("Configure Rails environment") {
+      govuk.setEnvar("RAILS_ENV", "test")
+    }
+
+    stage("Set up content schema dependency") {
```
suzannehamilton: Does this project rely on the content schemas? If not, you can remove this step, along with all the schema test setup:

- Parameter definition in lines 23-33
- Parameter initialization in lines 37-40
- Schema test branch build check in lines 42-44

mgrassotti: 👍 ok thanks

PR #107, opened 2017-02-02
Fix hash syntax for Organisation importer

chao-xian:
Following the `gds-api-adapters` version bump the hash syntax for the
organisation importer is wrong. This fixes both the importer and the
test.


chao-xian:
Thanks @NeilvB! 👍  (Cc @theleebriggs)

PR #108, opened 2017-02-02
Fix missing date_parser require

chao-xian:
We have some classes used by the rails app inside of lib. Controllers
explicitly require these classes but in this case the AnonymousFeedbackController
didn’t.

This was causing an error when running locally, but doesn’t surface on
deployed environments because they run in production mode which will
eager load.

In lieu of refactoring to put these classes in the app folder, this just
updates the controller to require the DateParser class.


NeilvB:
Looks good, thanks for fixing!

chao-xian:
Awesome! Thanks for checking @NeilvB (Cc @theleebriggs)

PR #109, opened 2017-05-08
Upgrade to rails 5

h-lame:
Best reviewed commit-by-commit to see the process and changes.  This is a minimal upgrade done without a full bundle update as I wanted to keep things as stable as possible.  I've also only upgraded to rails 5.0.2, not up to 5.1, again preferring to keep the change as small as possible.

For: https://trello.com/c/EiDlfUWI/159-migrate-to-rails-5-support-api


h-lame:
Currently broken because having upgraded to use the default `govuk.buildPipeline` we need to change that to support apps with `structure.sql` instead of `schema.rb`.  This is raised as htts://github.com/alphagov/govuk-puppet/pull/5886 and we should be able to re-run this build once that is merged.

PR #110, opened 2017-05-10
Provide rake task to fix misreported done page service feedback

h-lame:
For a couple of weeks in March 2017 the done page template sent the
service feedback form submissions to the support-api with the wrong
path - it used the slug of the service instead of the full path to
the done page (e.g. register-to-vote instead of /done/register-to-vote).
This meant that the service feedback was stored against the incorrect
url and couldn't be found in the feedback explorer for the done page.

Here we introduce an object that given a set of dates and a slug will
re-assign any service feedback (and aggregated service feedback) for
`/slug` to `/done/slug`.  There's also a version that will extract all
done pages that have some service feedback and fix all of them.

This feels safe to do because none of the /slug pages had a form on
them for submitting service feedback during the couple of weeks in
March so all service feedback in the db for /slug during that time
really is for /done/slug.  There are service feedback entries in the db
for some /slug pages outside those dates, but it's hard to know if those
are errors from similar bugs in the past, or from versions of the site
that did have service feedback forms on the /slug pages.  When we run
this script we'll limit it to 1st March 2017 to 1st April 2017.

For: https://trello.com/c/xZ97pseN/160-fix-data-in-feedex-for-done-pages-from-2nd-to-20th-march-2017

### DIFF
--- a/spec/lib/fix_misreported_done_page_service_feedback_spec.rb
+++ b/spec/lib/fix_misreported_done_page_service_feedback_spec.rb
@@ -1,0 +1,170 @@
+require 'fix_misreported_done_page_service_feedback'
+require 'rails_helper'
+
+RSpec.describe FixMisreportedDonePageServiceFeedback do
+  let(:start_date) { Date.new(2014, 6, 10) }
+  let(:inbetween_date) { start_date + 5 }
+  let(:end_date) { start_date + 10 }
+  let(:test_logger) { Rails.logger }
+
+  subject { described_class.new(start_date, end_date, test_logger) }
+
+  describe '#fix!' do
+    let(:service_slug) { 'register-to-vote' }
+    context 'for ServiceFeedback instances' do
+      context 'with "/done"-less version of supplied path' do
+        context 'created before the start date' do
+          let!(:service_feedback) do
+            FactoryGirl.create(
+              :service_feedback,
+              service_satisfaction_rating: 1,
+              slug: service_slug,
+              path: "/#{service_slug}",
+              created_at: start_date - 10,
+            )
+          end
+
+          it 'does not have its path changed' do
+            subject.fix!(service_slug)
+            expect(service_feedback.reload.path).not_to eq "/done/#{service_slug}"
+          end
+        end
+
+        context 'created after the end date' do
+          let!(:service_feedback) do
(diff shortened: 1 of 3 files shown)


spec/lib/fix_misreported_done_page_service_feedback_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+            FactoryGirl.create(
+              :service_feedback,
+              service_satisfaction_rating: 1,
+              slug: service_slug,
+              path: "/#{service_slug}",
+              created_at: inbetween_date,
+            )
+          end
+
+          it 'has its path change to the /done version' do
```
elenatanasoiu: changed*

spec/lib/fix_misreported_done_page_service_feedback_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+            FactoryGirl.create(
+              :aggregated_service_feedback,
+              service_satisfaction_rating: 1,
+              slug: service_slug,
+              path: "/#{service_slug}",
+              created_at: inbetween_date,
+            )
+          end
+
+          it 'has its path change to the /done version' do
```
elenatanasoiu: changed*

elenatanasoiu:
Needs rebasing as well

spec/lib/fix_misreported_done_page_service_feedback_spec.rb
elenatanasoiu: changed*

spec/lib/fix_misreported_done_page_service_feedback_spec.rb
elenatanasoiu: changed*

spec/lib/fix_misreported_done_page_service_feedback_spec.rb
h-lame: done in a rebase.

spec/lib/fix_misreported_done_page_service_feedback_spec.rb
h-lame: done in a rebase.

PR #118, opened 2017-10-27
Stop raising when no aggregate feedback

gpeng:
When aggregate feedback for a time period is empty we were raising an exception, seemingly to halt the flow and prevent the empty data being sent to performance platform. There was a test that checked that the error was swallowed when it was raised but not all callers of the code
were catching the error. This was causing it to be raised within the worker and then reported multiple times to Sentry when the job was retried.

This commit removes the exception and checks the `total` value from the payload instead using that to decide whether to send to performance platform or not.

[Trello](https://trello.com/c/btVExJP7/379-identify-and-fix-high-volume-sentry-errors)

### DIFF
--- a/app/workers/service_feedback_pp_uploader_worker.rb
+++ b/app/workers/service_feedback_pp_uploader_worker.rb
@@ -18,7 +18,9 @@ class ServiceFeedbackPPUploaderWorker
       slug: transaction_slug,
     ).call
 
-    api.submit_service_feedback_day_aggregate(transaction_slug, payload)
+    unless payload["total"].nil?
+      api.submit_service_feedback_day_aggregate(transaction_slug, payload)
+    end
   rescue GdsApi::PerformancePlatformDatasetNotConfigured => e
     Rails.logger.warn(e.message)
   end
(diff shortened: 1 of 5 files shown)


h-lame:
This is a good idea, aggregated data for a given path and day not existing is fundamentally an expected event so it seems unusual to raise an exception for it.  However, I wonder if we should log or statsd it instead?

gpeng:
Thanks @h-lame I've added an increment call in where the exception used to live. I'm not sure I see the use case for this data when we expect this to happen. Why do we need to log this? (or alternative question, am I recording the data you meant?)

h-lame:
I think I may have been being too conservative.  Now that I look back at this in more detail, I'm less convinced that statsd or logging is useful.  I thought the exception we raised had more info (like which path, which date, etc...) but it doesn't, and following that lead neither does the statsd increment.  We'd ultimately just have one number going up counting how many times we'd tried to run this against something that (for several good reasons) had no aggregated ratings.  I don't think this is useful.

There may be some other useful metrics or logging we could do around this, but as we basically ignore it now, I don't really see any reason to log or count it now either.  Sorry!  I think you can drop that latest commit with the statsd thing.  I'll approve on that assumption (I mean you can leave it in if you want too).

gpeng:
Excellent. Thanks. I've removed the statsd commit. :+1:

PR #121, opened 2017-12-07
Add search by format and by multiple paths

elenatanasoiu:
Add endpoints for searching for feedback by:
- multiple paths for feedback
- document type 

Also adds migration to populate the `document_type` for content items. 

For: https://trello.com/c/IB9P5EyE/266-introduce-new-browse-options-for-feedex

### DIFF
--- a/app/controllers/anonymous_feedback/document_types_controller.rb
+++ b/app/controllers/anonymous_feedback/document_types_controller.rb
@@ -1,0 +1,29 @@
+module AnonymousFeedback
+  class DocumentTypesController < ApplicationController
+    def index
+      render json: ContentItem.all_document_types
+    end
+
+    def show
+      ordering = if %w(path last_7_days last_30_days last_90_days).include? params[:ordering]
+                   params[:ordering]
+                 else
+                   'last_7_days'
+                 end
+
+      unless ContentItem.for_document_type(params[:document_type]).any?
+        head :not_found
+        return
+      end
+
+      anonymous_feedback_counts = ContentItem.
+          where(document_type: params[:document_type]).
+          summary(ordering)
+
+      render json: {
+          document_type: params[:document_type],
+          anonymous_feedback_counts: anonymous_feedback_counts,
+      }
+    end
+  end
+end
--- a/app/controllers/anonymous_feedback/export_requests_controller.rb
+++ b/app/controllers/anonymous_feedback/export_requests_controller.rb
@@ -28,7 +28,7 @@ class AnonymousFeedback::ExportRequestsController < ApplicationController
 
   private
     def export_request_params
-      permitted_params = [:from, :to, :organisation, :notification_email, :path_prefix, path_prefixes: []]
+      permitted_params = [:from, :to, :organisation, :notification_email, :document_type, :path_prefix, path_prefixes: []]
       clean_params = params.require(:export_request).permit(*permitted_params).to_h
       if clean_params[:path_prefix].present?
         clean_params[:path_prefixes] = [clean_params[:path_prefix]]
@@ -40,7 +40,8 @@ class AnonymousFeedback::ExportRequestsController < ApplicationController
           from: DateParser.parse(clean_params[:from]),
           to: DateParser.parse(clean_params[:to]),
           path_prefixes: clean_params[:path_prefixes],
-          organisation_slug: clean_params[:organisation]
+          organisation_slug: clean_params[:organisation],
+          document_type: clean_params[:document_type]
         },
         notification_email: clean_params[:notification_email]
       }
--- a/app/controllers/anonymous_feedback/export_requests_controller.rb
+++ b/app/controllers/anonymous_feedback/export_requests_controller.rb
@@ -1,3 +1,5 @@
+require 'date_parser'
+
 class AnonymousFeedback::ExportRequestsController < ApplicationController
 
   def create
@@ -26,13 +28,18 @@ class AnonymousFeedback::ExportRequestsController < ApplicationController
 
   private
     def export_request_params
-      permitted_params = %i(from to path_prefix organisation notification_email)
+      permitted_params = [:from, :to, :organisation, :notification_email, :path_prefix, path_prefixes: []]
       clean_params = params.require(:export_request).permit(*permitted_params).to_h
+      if clean_params[:path_prefix].present?
+        clean_params[:path_prefixes] = [clean_params[:path_prefix]]
+        clean_params.delete(:path_prefix)
+      end
+
       {
         filters: {
           from: DateParser.parse(clean_params[:from]),
           to: DateParser.parse(clean_params[:to]),
-          path_prefix: clean_params[:path_prefix],
+          path_prefixes: clean_params[:path_prefixes],
           organisation_slug: clean_params[:organisation]
         },
         notification_email: clean_params[:notification_email]
--- a/app/controllers/anonymous_feedback_controller.rb
+++ b/app/controllers/anonymous_feedback_controller.rb
@@ -4,7 +4,7 @@ class AnonymousFeedbackController < ApplicationController
   before_action :clean_params
 
   def index
-    unless params[:organisation_slug].present? || params[:path_prefixes].present?
+    unless at_least_one_param_present?
       head :bad_request
       return
     end
@@ -26,10 +26,15 @@ class AnonymousFeedbackController < ApplicationController
 
   private
 
+  def at_least_one_param_present?
+    params[:organisation_slug].present? || params[:path_prefixes].present? || params[:document_type].present?
+  end
+
   def scope
     AnonymousContact.
       for_query_parameters(path_prefixes: params[:path_prefixes],
                            organisation_slug: params[:organisation_slug],
+                           document_type: params[:document_type],
                            from: dates[0],
                            to: dates[1]).
       most_recent_first
--- a/app/controllers/anonymous_feedback_controller.rb
+++ b/app/controllers/anonymous_feedback_controller.rb
@@ -1,8 +1,10 @@
 require 'date_parser'
 
 class AnonymousFeedbackController < ApplicationController
+  before_action :clean_params
+
   def index
-    unless params[:organisation_slug].present? || params[:path_prefix].present?
+    unless params[:organisation_slug].present? || params[:path_prefixes].present?
       head :bad_request
       return
     end
@@ -26,11 +28,15 @@ class AnonymousFeedbackController < ApplicationController
 
   def scope
     AnonymousContact.
-      for_query_parameters(path_prefix: params[:path_prefix],
+      for_query_parameters(path_prefixes: params[:path_prefixes],
                            organisation_slug: params[:organisation_slug],
                            from: dates[0],
                            to: dates[1]).
       most_recent_first
+  end
+
+  def clean_params
+    params[:path_prefixes] = [params[:path_prefix]] if params[:path_prefix].present?
   end
 
   def dates
--- a/app/models/anonymous_contact.rb
+++ b/app/models/anonymous_contact.rb
@@ -24,6 +24,7 @@ class AnonymousContact < ApplicationRecord
   scope :most_recent_last, -> { order("created_at ASC") }
   scope :created_between_days, -> (first_date, last_date) { where(created_at: first_date..last_date.at_end_of_day) }
   scope :for_organisation_slug, -> (slug) { joins(:organisations).where(organisations: {slug: slug}) }
+  scope :for_document_type, ->(document_type) { joins(:content_item).where(content_items: { document_type: document_type }) }
 
   scope :matching_path_prefixes, ->(paths) do
     if paths.present?
@@ -37,6 +38,7 @@ class AnonymousContact < ApplicationRecord
     from = options[:from] || Date.new(1970)
     to = options[:to] || Date.today
     organisation_slug = options[:organisation_slug]
+    document_type = options[:document_type]
 
     query = only_actionable.
       free_of_personal_info.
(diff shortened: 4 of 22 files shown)


app/controllers/anonymous_feedback_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 class AnonymousFeedbackController < ApplicationController
   def index
-    unless params[:organisation_slug].present? || params[:path_prefixes].present?
+    unless params[:organisation_slug].present? || params[:path_prefixes].present? || params[:document_type].present?
```
h-lame: With three possible options I kinda feel like it might be worth extracting this to a method and giving it a name so it's easier to read the action.

spec/controllers/anonymous_feedback/export_requests_controller_spec.rb
```diff
@@ -7,8 +7,12 @@
...
         post :create, params: {
           export_request: {
-            from: "2015-05-01", to: "2015-06-01",
-            path_prefixes: ["/"], notification_email: "foo@example.com"
+            from: "2015-05-01",
+            to: "2015-06-01",
+            path_prefixes: ["/"],
+            notification_email: "foo@example.com",
+            organisation: "",
+            document_type: ""
```
h-lame: :+1: for reformatting this

app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -1,3 +1,5 @@
+require 'date_parser'
```
h-lame: The commit message here says "a comma separated list of paths needs to be uploaded in the “By URL List” section" but that's refereing to a UI in a different app.  I think we should reword to only talk about how to submit a multiple path search using the API (e.g. by suppling extra paths in the params).

app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -26,13 +28,13 @@ def show
...
     def export_request_params
-      permitted_params = %i(from to path_prefix organisation notification_email)
+      permitted_params = [:from, :to, :organisation, :notification_email, path_prefixes: []]
       clean_params = params.require(:export_request).permit(*permitted_params).to_h
       {
         filters: {
           from: DateParser.parse(clean_params[:from]),
           to: DateParser.parse(clean_params[:to]),
-          path_prefix: clean_params[:path_prefix],
+          path_prefixes: clean_params[:path_prefixes],
```
h-lame: I think we might need to support `path_prefix` as a param for a while yet (and we "upgrade" it internally to `path_prefixes: [path_prefix_value]`).  Otherwise we have to be very carefuly about synchronising deploys so we don't break support by shipping a version of support-api that doesn't understand `path_prefix` anymore, or vice-versa by shipping a version of support that sends `path_prefixes` but support-api doesn't understand that yet.

app/models/anonymous_contact.rb
```diff
@@ -22,12 +22,18 @@ class AnonymousContact < ApplicationRecord
...
   scope :most_recent_first, -> { order("created_at DESC") }
   scope :most_recent_last, -> { order("created_at ASC") }
-  scope :matching_path_prefix, ->(path) { where("anonymous_contacts.path LIKE ?", path + "%") if path }
   scope :created_between_days, -> (first_date, last_date) { where(created_at: first_date..last_date.at_end_of_day) }
   scope :for_organisation_slug, -> (slug) { joins(:organisations).where(organisations: {slug: slug}) }
 
+  scope :matching_path_prefixes, ->(paths) do
+    if paths.present?
+      similar_to = paths.map { |p| "'#{p}%'" }.join(", ")
+      where("anonymous_contacts.path LIKE ANY (ARRAY[#{similar_to}])")
```
h-lame: Ooooh.  Nice.

app/models/anonymous_contact.rb
```diff
@@ -22,12 +22,18 @@ class AnonymousContact < ApplicationRecord
   scope :only_actionable, -> { where(is_actionable: true) }
   scope :most_recent_first, -> { order("created_at DESC") }
   scope :most_recent_last, -> { order("created_at ASC") }
-  scope :matching_path_prefix, ->(path) { where("anonymous_contacts.path LIKE ?", path + "%") if path }
   scope :created_between_days, -> (first_date, last_date) { where(created_at: first_date..last_date.at_end_of_day) }
   scope :for_organisation_slug, -> (slug) { joins(:organisations).where(organisations: {slug: slug}) }
 
+  scope :matching_path_prefixes, ->(paths) do
+    if paths.present?
+      similar_to = paths.map { |p| "'#{p}%'" }.join(", ")
```
h-lame: do we open ourselves up to any potential SQL injection here?

app/models/feedback_export_request.rb
```diff
@@ -17,7 +17,8 @@ def generate_filename!
       filters[:from].nil? ? "0000-00-00" : filters[:from].to_date.iso8601,
       filters[:to].nil? ? Date.today.iso8601 : filters[:to].to_date.iso8601,
     ]
-    parts += filters[:path_prefix].split("/").reject(&:blank?) if filters[:path_prefix].present?
+
+    parts += filters[:path_prefixes].first.split("/").reject(&:blank?) if filters[:path_prefixes].present?
```
h-lame: So I think the previous version of this line stripped all slashes from the path and then added the path segments (the reject blank bit is to ignore the leading and trailing spaces) to the bits of the filename.  This new version is going to do that for the first path only, should we do it for all of them?  Or add some indicator that there are other paths like:

```ruby
if filters[:path_prefixes].present?
  parts += path_segments_of_first_path
  parts += ['and', filters[:path_prefixes].count - 1, 'other', 'paths'] if filters[:path_prefixes].size > 1
end
```

Something like that, but with better pluralisation checking.  Maybe?

spec/requests/anonymous_feedback_feed_spec.rb
```diff
@@ -41,7 +41,7 @@
     end
 
     it "returns feedback with the appropriate fields in reverse chronological order" do
-      get_json "/anonymous-feedback?path_prefix=/"
+      get_json "/anonymous-feedback?path_prefixes[]=/"
```
h-lame: I feel like this would benefit another handful of tests that actually show us doing searches with more than one path. At the moment the only test that actually exercises multiple paths is a scope test on the anonymous contact model.  I'd expect to see some in the controllers, maybe an integration test, and some on the export filename generator too.

spec/controllers/anonymous_feedback/document_types_controller_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+        expect(JSON.parse(response.body)).to be_eql(%w[case_study smart_answer])
+      end
+
+      it 'filters out nils' do
+        expect(JSON.parse(response.body)).to_not include(nil)
+      end
+
+      it 'filters out blank document types' do
+        expect(JSON.parse(response.body)).to_not include("")
+      end
```
h-lame: :+1:

spec/controllers/anonymous_feedback/document_types_controller_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+            'path' => '/search',
+            'last_7_days' => 2,
+            'last_30_days' => 2,
+            'last_90_days' => 2
+        }]
+        )
+      end
+    end
+  end
+end
```
h-lame: Indentation in this spec is a bit wonky, I suspect rubocop has "fixed" some stuff and we need to fix it manually.  Rubocop seems to have problems when you put `[{` on one line and then indent, so maybe it makes sense to add more newlines between `[` and `{`.

spec/models/anonymous_contact_spec.rb
```diff
@@ -180,6 +203,28 @@ def contact(options = {})
...
+        let(:document_type) { "smart_answer" }
+
+        it { is_expected.to eq([contact_1, contact_2]) }
+      end
+
+      context "with a non-existent document type" do
+        let(:document_type) { "made-up-document-type" }
+
+        it { is_expected.to eq([contact_1, contact_2]) }
+      end
```
h-lame: This seems like an odd result.  Why does using a made-up document-type filter the results the same way as when I provide "smart-answers"?  I'd expect it would return nothing (because nothing matches that document type) or return everything (because we ignore the option entirely).

elenatanasoiu: It does a join with content_items here:
` scope :for_document_type, ->(document_type) { joins(:content_item).where(content_items: { document_type: document_type })}`

Because it won't find the "made-up-document-type" in the `content_items` table, it will return ALL the contacts THAT RESULT FROM THE JOIN. 

In our case these are `contact_1` and `contact_2`, because `contact_3` does not have a corresponding content item so will be ignored. 

Ultimately what you're searching for are anonymous contacts that DO correspond to a content_item, otherwise they wouldn't be relevant so I think it's returning the correct result: all contacts that are actually tied to a document type, orphans are ignored.

Query is here: 
https://github.com/alphagov/support-api/pull/121/files#diff-a89082f91c7d60c944440b15a86d9019R27

spec/controllers/anonymous_feedback/document_types_controller_spec.rb
elenatanasoiu: Hope this has been fixed. I had to fight Rubocop for a bit with this.

spec/models/anonymous_contact_spec.rb
h-lame: On closer inspection the problem here is on [line 170 above](https://github.com/alphagov/support-api/pull/121/files#diff-0c3cf3646efeddc59b33510b319b6eb1R170) - we're not passing the document_type into the query used by the test.  The where clause should be reducing the result set to nothing, but it's not because we're always using "smart-answer".

elenatanasoiu: :+1: corrected

app/controllers/anonymous_feedback/document_types_controller.rb
```diff
@@ -0,0 +1,29 @@
+module AnonymousFeedback
+  class DocumentTypesController < ApplicationController
+    def index
+      render json: ContentItem.all_document_types
+    end
+
+    def show
+      ordering = if %w(name last_7_days last_30_days last_90_days).include? params[:ordering]
```
h-lame: My bad, I asked you to change this to name, but path was absolutely the right thing to do. I misunderstood what the data in these responses actually is.  It's a list of paths that received feedback with counts for the last 7, 30, and 90 days.  And the document type filters out which paths.  I thought it was a list of document types, but it's not.

Sorry!

elenatanasoiu: Changed back to path. I should've double checked. 👍

spec/models/feedback_export_request_spec.rb
```diff
@@ -35,8 +35,7 @@
                               from: from,
                               to: to,
                               organisation_slug: organisation_slug
-                            }
-                         )
+                            })
```
h-lame: If rubocop said you had to fix this then I think:

```ruby
let(:instance) do
  described_class.new(
    filters: {
      path_prefix: path_prefix,
      path_prefixes: path_prefixes,
      from: from,
      to: to,
      organisation_slug: organisation_slug
    }
  )
end
```

would be clearer.

spec/controllers/anonymous_feedback_controller_spec.rb
```diff
@@ -179,6 +200,52 @@
       end
     end
 
+    describe "filter by document type" do
+      let(:hmrc) { create(:organisation, slug: 'hm-revenue-customs') }
+      let(:smart_answer) {
+        create(:content_item,
+                                  document_type: 'smart_answer',
```
h-lame: Can probably pull these args in a bit.  Maybe rubocop would allow:

```ruby
create(
  :content_item,
  document_type: 'smart_answer',
  ...
)
```

spec/controllers/anonymous_feedback_controller_spec.rb
```diff
@@ -96,6 +96,27 @@
...
+        get :index, params: { path_prefixes: ["/non-existent"] }
+
+        expect(response).to be_success
+        expect(json_response).to eq(
+          "results" => [],
+          "page_size" => 50,
+          "total_count" => 0,
+          "current_page" => 1,
+          "pages" => 0,
+                                     )
```
h-lame: This can be pulled right in to line up with the `expect` above.

spec/models/feedback_export_request_spec.rb
```diff
@@ -210,8 +209,7 @@
                               to: Date.new(2015, 5),
                               path_prefixes: ["/"],
                               organisation_slug: "hm-revenue-customs"
-                            }
-      ).results
+                            }).results
```
h-lame: I think we can do the same thing I suggested above here too.

spec/controllers/anonymous_feedback_controller_spec.rb
elenatanasoiu: Re-aligned

spec/controllers/anonymous_feedback_controller_spec.rb
elenatanasoiu: Re-aligned

spec/models/feedback_export_request_spec.rb
elenatanasoiu: Reformatted

spec/models/feedback_export_request_spec.rb
elenatanasoiu: Restyled

PR #174, opened 2018-02-26
Add endpoint to retrieve the count of feedback by day

peteglondon:
content-performance-manager needs to retrieve metrics on the number of comment for each content item by day in order to populate it's data warehouse.

This PR implements a new endpoint /feedback-by-day that exposes this information.

the Trello card is [Expose Feedex numbers via Support API](https://trello.com/c/6nz6zXEE/113-5-expose-feedex-numbers-via-support-api)

### DIFF
--- a/app/controllers/feedback_by_day_controller.rb
+++ b/app/controllers/feedback_by_day_controller.rb
@@ -1,15 +1,24 @@
 require 'date_parser'
 
 class FeedbackByDayController < ApplicationController
+  before_action :validate_params!, only: :index
+
   def index
-    unless date
-      head :bad_request
-      return
-    end
-    render json: {results: FeedbackByDay.retrieve(date)}
+    render json: FeedbackByDay.retrieve(date, params[:page], params[:per_page])
   end
 
 private
+
+  def validate_params!
+    head :bad_request unless valid_params?
+  end
+
+  def valid_params?
+    return false if params[:page] && !params[:page].match(/^\d+$/)
+    return false if params[:per_page] && !params[:per_page].match(/^\d+$/)
+    date
+  end
+
   def date
     DateParser.parse(params[:date])
   end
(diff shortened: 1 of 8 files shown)


app/models/feedback_by_day.rb
```diff
@@ -0,0 +1,37 @@
...
+      page_size: summary_query.limit_value
+    }
+  end
+
+private
+
+  def summary_query
+    AnonymousContact.where(created_at: get_range)
+      .group(:path).page(current_page).per(per_page)
+      .order(:path)
```
pmanrubia: We don't really need to sort by path. It will speed up queries if we don't sort.

spec/models/feedback_by_day_spec.rb
```diff
@@ -0,0 +1,103 @@
+require 'rails_helper'
+require 'feedback_by_day'
+
+describe FeedbackByDay do
+  before :each do
+    create_list(:problem_report, 5, path: '/browse/abroad', created_at: Time.utc(2018, 0o2, 1, 0, 1, 0))
```
pmanrubia: Why are we using `0o2`?

app/models/feedback_by_day.rb
peteglondon: It makes it really hard to test the pagination if we don't sort by something.

spec/models/feedback_by_day_spec.rb
peteglondon: Sorry - Typo - fixed now

app/controllers/feedback_by_day_controller.rb
```diff
@@ -0,0 +1,23 @@
+require 'date_parser'
+
+class FeedbackByDayController < ApplicationController
+  def index
```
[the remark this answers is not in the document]
peteglondon: done

PR #237, opened 2018-09-06
Add an index on organisation_id to content_item_organisations

thomasleese:
This makes a query which is currently taking around 20 seconds to run, complete in 320ms.

Found by looking in the Postgres logs for slow running queries.

[Trello Card](https://trello.com/c/VztlS695/447-investigate-options-for-reducing-load-on-the-production-postgres-primary)

### DIFF
--- a/db/migrate/20180906145408_add_index_to_content_items_organisations_organisation_id.rb
+++ b/db/migrate/20180906145408_add_index_to_content_items_organisations_organisation_id.rb
@@ -1,0 +1,5 @@
+class AddIndexToContentItemsOrganisationsOrganisationId < ActiveRecord::Migration[5.0]
+  def change
+    add_index :content_items_organisations, :organisation_id
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20180906145408_add_index_to_content_items_organisations_organisation_id.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToContentItemsOrganisationsOrganisationId < ActiveRecord::Migration[5.0]
+  def change
+    add_index :content_items_organisations, :organisation_id
```
barrucadu: Do we want to create this index concurrently?

thomasleese: It took under a second for the index to be created in staging, so I figured it wouldn't be an issue. I can look into it though.

thomasleese:
@barrucadu I've sadly removed the commit which changes it to a `schema.rb`.

PR #263, opened 2018-12-18
Move CSV exports to S3

rubenarakelyan:
This commit changes the support-api app to save CSV exports to S3.

Depends on https://github.com/alphagov/support/pull/539


rubenarakelyan:
No, there doesn't seem to be any retention policy and there are files in there from 2015, but I will look into setting it up since it doesn't make sense to keep them there.

PR #278, opened 2019-01-29
Add endpoint to capture business finder feedback

peteglondon:
Trello: https://trello.com/c/a9WXTd6f/64-build-a-feedback-component-to-sit-below-business-finder-results

# What
Add and endpoint to capture feedback from the EU Exit business finder.

# Why

We want to build a feedback component that sits at the bottom of the business finder results that allows users to tell us what's missing. 

There is a [related PR in feedback](https://github.com/alphagov/feedback/pull/630) that sends the feedback to support API

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ gem 'govuk_app_config', '~> 1.11.2'
 if ENV['API_DEV']
   gem "gds-api-adapters", :path => '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '~> 57.2'
+  gem 'gds-api-adapters', '~> 57.3.1'
 end
 
 gem 'kaminari', '1.1.1'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -94,7 +94,7 @@ GEM
       fog-core
       nokogiri (>= 1.5.11, < 2.0.0)
     formatador (0.2.5)
-    gds-api-adapters (57.2.3)
+    gds-api-adapters (57.3.1)
       addressable
       link_header
       null_logger
@@ -385,7 +385,7 @@ DEPENDENCIES
   climate_control (~> 0.2.0)
   factory_bot_rails (~> 4.11.1)
   fog-aws (~> 3.3)
-  gds-api-adapters (~> 57.2)
+  gds-api-adapters (~> 57.3.1)
   gds-sso (~> 14.0.0)
   gds_zendesk (= 3.0.0)
   govuk-lint
--- a/app/controllers/anonymous_feedback/content_improvement_controller.rb
+++ b/app/controllers/anonymous_feedback/content_improvement_controller.rb
@@ -1,0 +1,19 @@
+module AnonymousFeedback
+  class ContentImprovementController < ApplicationController
+    def create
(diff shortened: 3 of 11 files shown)


peteglondon:
[Build a feedback component to sit below Business Finder results](https://trello.com/c/a9WXTd6f/64-build-a-feedback-component-to-sit-below-business-finder-results)

app/controllers/anonymous_feedback/business_finder_controller.rb
```diff
@@ -0,0 +1,19 @@
+module AnonymousFeedback
+  class BusinessFinderController < ApplicationController
```
steventux: Is there any chance of this being used in a more generic context? ie. is it conceivable we'd want to re-use a content-specific feedback mechanism elsewhere? The params passed to the feedback mechanism aren't particularly specific to the finder in question, they could conceivably come from anywhere. 
Would it be better to call it something less single-use and use a key or descriptor to signify the origin of the feedback? 
Also I'd argue we're baking knowledge of a client app into the support-api by lacing it with _business-finder_ nomenclature, the support-api handles feedback, it probably shouldn't be wired too directly to what provides the feedback.
How about `ContentImprovementFeedback` or something dealing specifically with suggestions to improve content? We could store a complete URL or path+query string to denote the context of the feedback.

app/controllers/anonymous_feedback/business_finder_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+        BusinessFinderFeedbackWorker.perform_async(bus_finder_feedback_params)
+        head :accepted
+      else
+        render json: { "errors" => feedback.errors.to_a }, status: 422
+      end
+    end
+
+  private
+
+    def bus_finder_feedback_params
```
steventux: Suggest `feedback_params`

db/migrate/20190128164439_create_business_finder_feedbacks.rb
```diff
@@ -0,0 +1,10 @@
+class CreateBusinessFinderFeedbacks < ActiveRecord::Migration[5.2]
+  def change
+    create_table :business_finder_feedbacks do |t|
```
steventux: I think we could get more use out of this by making it a generic mechanism for suggesting content improvement by storing a path with req params against the feedback, this would show the exact set of results in the case of a finder but would also allow us to add content suggestion components to other pages.

app/models/business_finder_feedback.rb
```diff
@@ -0,0 +1,17 @@
+require 'field_which_may_contain_personal_information'
+
+class BusinessFinderFeedback < ApplicationRecord
```
steventux: The finder doesn't find businesses, this might not mean much to developers in future even though we all understand what it means right now. I think something less specific might be more suitable.

app/workers/business_finder_feedback_worker.rb
```diff
@@ -0,0 +1,7 @@
+class BusinessFinderFeedbackWorker
+  include Sidekiq::Worker
+
+  def perform(bus_finder_feedback_params)
```
steventux: no buses were found :bus: I'd change this to `feedback_params` the support API shouldn't care about the origin like this.

PR #304, opened 2019-04-15
Specify ActiveRecord::Migration[4.2] for all existing migrations

issyl0:
- We were seeing errors in Sentry related to these not being specified:
  https://sentry.io/organizations/govuk/issues/971865094/?project=202256&query=is%3Aunresolved+environment%3Astaging


cbaines:
I'm guessing this is because you're attempting to migrate from an empty database.

This problem has occurred previously during deploys to Staging, where the database has been empty due to issues with the data sync process from Production.

It's probably better to populate the database with data, rather than make the historic migrations work, as a database with data is easier to use for testing in Staging.

issyl0:
Yep. I was just trying to get _something_ working. But there's definitely a problem with the data sync. Let's investigate that more rather than dealing with a symptom of it.

PR #308, opened 2019-04-18
Add a Rake task to delete old Zendesk tickets and users

issyl0:
- We have a need to delete old Zendesk tickets and users, for members of
  the public who haven't contacted us in a year, for GDPR reasons.
- This adds a Rake task which calls a script in `lib/`.
- Users are created in Zendesk for every contact form submission that
  people send us. That means there are a lot (~2 million users).
- This uses the zendesk_api gem, the officially supported Zendesk gem.
  This app also uses the gds_zendesk gem, our own wrapper around _some of
  the functionality_, but it doesn't support deletion or more recent
  Zendesk API features, hence using the official one for this process.

Considerations:
- This script does not have any tests. The effort involved in stubbing the Zendesk API calls we use seemed to outweigh the benefits, and there's no way to create tickets with `created_at` dates in the past.

https://trello.com/c/MaMoUari/350-spike-script-to-delete-users-from-zendesk

Co-authored-by: Aditya Pahuja <aditya.pahuja@digital.cabinet-office.gov.uk>
Co-authored-by: Andy Paine <andy.paine@digital.cabinet-office.gov.uk>
Co-authored-by: Athanasios Voutsadakis <athanasios.voutsadakis@digital.cabinet-office.gov.uk>


lib/zendesk_cleaner.rb
```diff
@@ -0,0 +1,75 @@
...
+  def delete_old_users
+    @client.users.all! do |user|
+      if is_end_user?(user) &&
+          @users_with_existing_tickets.exclude?(user.id) &&
+          older_than_one_year?(user) &&
+          has_not_logged_in_within_a_year?(user)
+
+        @number_of_users_deleted += 1
+        puts "Num of deleted users #{@number_of_users_deleted}"
+        #user.destroy!
```
issyl0: We commented out the actual destroy commands because we didn’t want to _actually_ delete all the things when we were testing refactoring the script. We confirmed that they worked before though.

lib/zendesk_cleaner.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def older_than_one_year?(user)
+    user.created_at < 1.year.ago.to_datetime
+  end
+
+  def delete_tickets_older_than_a_year(ticket)
+    if not_updated_for_a_year?(ticket)
+      @number_of_tickets_deleted += 1
+      puts "Num of deleted tickets #{@number_of_tickets_deleted}"
+      #ticket.destroy!
```
issyl0: As above re: commenting.

lib/zendesk_cleaner.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def initialize(client)
+    @client = client
+    @users_with_existing_tickets = Set.new
+    @number_of_tickets_deleted = 0
+    @number_of_users_deleted = 0
+  end
+
+  def display_current_num_of_tickets_and_users
+    puts "The sandbox account has #{@client.users.count} users and #{@client.tickets.count} tickets."
```
thomasleese: Since you have the `attr_accessor` above (which might be better as an `attr_reader`) you don't need the `@` in front of all your instances variables here and later in the class.

issyl0:
I'm off on holiday for a week, and Aditya is off for a month. I've handed this over to @andy-paine.

issyl0:
We (Autom8) about to run this for real (on a box of ours that won't turn off, to avoid clogging up Jenkins for ~24h!), so I've marked this as ready for review as it feels weird running code from a "draft" PR.

issyl0:
I'm going to close this PR. We've migrated these scripts to https://github.com/alphagov/zendesk-scripts, as this lot of code has outgrown the scope of `support-api`. It was a good idea in the beginning when we thought we could use the `gds-zendesk` gem. Thanks for the initial review, @thomasleese, and to everyone else for suitably ignoring this/letting this stay open for so long!

PR #336, opened 2019-07-02
Use environment variables to configure database

tijmenb:
This will make using this app in govuk-docker simpler. I've copied this config from [publishing-api][1].

[1]: https://github.com/alphagov/publishing-api/blob/a10afeb4790f737274d83a710208f64016fc60b9/config/database.yml

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -10,8 +10,14 @@ default: &default
 
 development:
   <<: *default
-  database: support_contacts_development
+  database: support_api_development
+  url: <%= ENV["DATABASE_URL"]%>
 
 test:
   <<: *default
-  database: support_contacts_test
+  database: support_api_test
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -10,8 +10,14 @@ default: &default
...
   <<: *default
-  database: support_contacts_development
+  database: support_api_development
+  url: <%= ENV["DATABASE_URL"]%>
 
 test:
   <<: *default
-  database: support_contacts_test
+  database: support_api_test
+  url: <%= ENV["DATABASE_URL"].try(:sub, /([-_]development)?$/, '_test')%>
```
kevindew: It'd be better to use an explicit env var here of TEST_DATABASE_URL. That was the revised approach, that came post Publishing API, since DATABASE_URL env var doesn't necessarily mean development and you can end up with test using same DB pretty easy.

PR #411, opened 2020-01-31
Upload exported CSV file to S3 rather than attaching it

benjamineskola:
Use S3 to transfer files, to be consistent with other emails and also compatible (in future) with Notify.

https://trello.com/c/V9wtOj6x/1713-3-support-api-make-globalcsvexport-use-s3-instead-of-email-attachments

### DIFF
--- a/app/mailers/global_export_notification.rb
+++ b/app/mailers/global_export_notification.rb
@@ -1,10 +1,9 @@
 class GlobalExportNotification < ApplicationMailer
-  def notification_email(notification_email, filename, csv_contents)
-    attachments[filename] = csv_contents
-
+  def notification_email(notification_email, url)
+    @url = url
     mail(
       to: notification_email,
-      subject: "[GOV.UK Feedback Explorer] Your global export is attached",
+      subject: "[GOV.UK Feedback Explorer] Your global export is ready",
     )
   end
 end
--- a/app/views/global_export_notification/notification_email.text.erb
+++ b/app/views/global_export_notification/notification_email.text.erb
@@ -2,6 +2,6 @@ Hi,
 
 Use the link below to download your CSV file from Feedback Explorer.
 
-- Feedex
-
 <%= @url %>
+
+- Feedex
--- a/app/views/global_export_notification/notification_email.text.erb
+++ b/app/views/global_export_notification/notification_email.text.erb
@@ -1,5 +1,7 @@
 Hi,
 
-Your CSV file is attached.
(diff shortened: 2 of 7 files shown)


app/mailers/global_export_notification.rb
```diff
@@ -1,10 +1,9 @@
 class GlobalExportNotification < ApplicationMailer
-  def notification_email(notification_email, filename, csv_contents)
-    attachments[filename] = csv_contents
-
+  def notification_email(notification_email, url)
+    @url = url
     mail(
       to: notification_email,
-      subject: "[GOV.UK Feedback Explorer] Your global export is attached",
+      subject: '[GOV.UK Feedback Explorer] Your global export is attached'
```
nicholsj: A small thing, but "attached" might confuse users who would expect an attachment and not get one. How about `Your global export is ready`?

benjamineskola: I was just wondering what if anything to change that to.

app/workers/generate_global_export_csv_worker.rb
```diff
@@ -8,6 +8,24 @@ def perform(export_params)
       export_params["exclude_spam"],
     ).call
 
-    GlobalExportNotification.notification_email(export_params["notification_email"], filename, contents).deliver_now
+    s3_file = self.class.save_file_to_s3(filename, contents)
+
+    GlobalExportNotification.notification_email(export_params["notification_email"], s3_file.key).deliver_now
+  end
+
+  def self.save_file_to_s3(filename, csv)
```
thomasleese: This function is the same as https://github.com/alphagov/support-api/blob/cad664e02c54fd48861b03389ba01ce8f6b6e7c0/app/workers/generate_feedback_csv_worker.rb#L20-L34, could we extract it out into a service class?

app/views/global_export_notification/notification_email.text.erb
```diff
@@ -1,5 +1,7 @@
 Hi,
 
-Your CSV file is attached.
+Use the link below to download your CSV file from Feedback Explorer.
 
 - Feedex
+
+<%= @url %>
```
thomasleese: I _think_ this belongs above `- Feedex`, I think that's supposed to be a "from Feedex" sort of line, but I'm not sure what benefit there is even having it.

benjamineskola: It's a strange one; the other template didn't have it (which I copied the body text from).

PR #413, opened 2020-02-05
Revert "Upload exported CSV file to S3 rather than attaching it"

benjamineskola:
Reverts alphagov/support-api#411 which introduces a bug; corrected PR forthcoming.


thomasleese:
I agree it might have been easier to fix forward here, Support API doesn't get released very often, and it's possible to add messages in the Release app telling other people not to deploy it, but oh well!

PR #424, opened 2020-02-21
Use GOV.UK Notify for ActionMailer messages

benjamineskola:
We want to switch to using GOV.UK Notify instead of Amazon SES for emails. This instantiates the Notify client and configures ActionMailer to use it.

https://trello.com/c/GZ142gQv/1730-5-support-api-use-notify-instead-of-amazon-ses-%F0%9F%8D%90

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -20,3 +20,4 @@ test:
 # instead read values from the environment.
 production:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
+  notify_api_key: <%= ENV["GOVUK_NOTIFY_API_KEY"] %>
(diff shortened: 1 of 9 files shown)


issyl0:
```
+ bundle exec govuk-lint-ruby --parallel --format html --out rubocop-6b001e1.html --format clang app lib spec test
[DEPRECATION] `govuk-lint-ruby` is deprecated.  Please use `rubocop` with `rubocop-govuk` instead.
```

Oh no. You made linting changes to appease an old version of a deprecated linter.

config/secrets.yml
```diff
@@ -20,3 +20,4 @@ test:
 # instead read values from the environment.
 production:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
+  notify_api_key: <%= ENV['GOVUK_NOTIFY_API_KEY'] %>
```
issyl0: We'll need to do some Puppeteering with some secretsing as well.

PR #492, opened 2020-07-27
Make app consistent with upgrades

cdccollins:
This is to make Support API consistent with the other apps that have been upgraded to Rails 6 - specifically layout of Gemfile and updating the gems.


EGiataganas:
Hey @cdccollins this looks great, I've compared the PR changes with [railsdiff][1] and there are some changes that could also be done:

1. Remove `bin/bundle`, `bin/update`
2. Update [system! "bin/rails db:setup"][2] with `system! "bin/rails db:prepare"`
3. In the `config/application.rb` remove the unused (commented) Railties require statements https://github.com/alphagov/travel-advice-publisher/pull/919/commits/c59db4c22030b33b458e0b88e91c9f807b7ec123
4. Check whether we need to explicitly add the local dev link http://support-api.dev.gov.uk/ to the config hosts in the `config/environments/development.rb` https://github.com/alphagov/travel-advice-publisher/pull/919/commits/2689aa4
5. In the `config/environments/test.rb` the `config.cache_classes = true` has changed to `false` and there is `config.action_view.cache_template_loading = true`
6. Remove `config/initializers/new_framework_defaults.rb` as per the reason in https://github.com/alphagov/travel-advice-publisher/commit/8b0745e
7. In the `config/spring.rb` the `%w[]` needs to be updated to `Spring.watch()`
```ruby
Spring.watch(
  ".ruby-version",
  ".rbenv-vars",
  "tmp/restart.txt",
  "tmp/caching-dev.txt"
)
```
8. Add the Rails version number to the top 27 migrations [5.0] as otherwise a new migration will fail

[1]: http://railsdiff.org/5.2.4.3/6.0.3.2
[2]: https://github.com/alphagov/support-api/blob/master/bin/setup#L27

cdccollins:
> ```ruby
> Spring.watch(
>   ".ruby-version",
>   ".rbenv-vars",
>   "tmp/restart.txt",
>   "tmp/caching-dev.txt"
> )
> ```

cdccollins:
> 6. Remove `config/initializers/new_framework_defaults.rb` as per the reason in [alphagov/travel-advice-publisher@8b0745e](https://github.com/alphagov/travel-advice-publisher/commit/8b0745e)

If I remove it, some of the tests fail. And as pudim pointed out in https://github.com/alphagov/frontend/pull/2436, they may be needed in production but not in the tests.

PR #541, opened 2021-03-17
Revise tech docs in README to favour GOV.UK Docker

benthorner:
https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This follows the pattern established in [1].

Please see the commit messages for more details.

[1]: alphagov/content-publisher#2257

PR #542, opened 2021-03-17
Remove code related to performance platform

thomasleese:
The performance platform has been shut down, so these pieces of code will stop working correctly. We don't have a suitable alternative set up, so we can remove the functionality from this app.


richardTowers (approved):

benthorner:
This relates to the PR I've got to rewrite the README for this app: https://github.com/alphagov/support-api/pull/541

> Support API also stores ratings for services. It aggregates those ratings every day and writes that data
to the [performance platform][performance-platform] write API.  We can also use the performance platform
read API to import those ratings if there's ever a discrepancy.

I notice we still have the aggregate rating code in the app. Is that intentional? I'm wondering what to say about the rating data - perhaps that it's a legacy feature that we may use again one day?

thomasleese:
> I notice we still have the aggregate rating code in the app. Is that intentional? I'm wondering what to say about the rating data - perhaps that it's a legacy feature that we may use again one day?

I only wanted to remove the code which would actually break (and therefore result in Sentry errors, high retry queues, etc) as a result of the Performance Platform being shut down. I think it's a reasonable question to ask how far we go with removing these, but it feels like it needs some wider thought. I'll have a think about whether this is something Platform Health wants to take on.

Alternatively, if the only bit left is the aggregate rating code, then I might just remove it since it's not useful without the code that sends it to the Performance Platform.

benthorner:
Thanks. I had a quick look, and found:

- There's an additional file that can be removed: `app/models/corporate_content_problem_report_aggregated_metrics.rb`
- It's necessary to keep the [aggregator](https://github.com/alphagov/support-api/blob/3b6a83419f9c9969726c9488f599e45fb46ec279/lib/service_feedback_aggregator.rb#L13), as it does some DB health work as well (and we may want the stats in future).

I've amended my [README PR](https://github.com/alphagov/support-api/pull/541) to explain the situation, would you like to review it?

(If you want to remove that other file, I can review that PR.)

PR #595, opened 2021-12-07
Switch from a db/structure.sql to db/schema.rb

kevindew:
The change to a structure.sql was made back in 2015 [1], to make use of
a feature ActiveRecord didn't support: varchar_pattern_ops.

This is now supported by ActiveRecord so we can return to the Rails
default.

I tested there weren't problems by dumping and recreating the database
and comparing to structure.sql, the differences were negligible.

[1]: https://github.com/alphagov/support-api/commit/797b41f9281da5b7e7a1b7f108528bbd0d7ef7ef#diff-1dd5a8f580b9615769ddc19a64a98f455d9d20eb149c443f41d9d3c797d39b12


hannako (approved):
👍

PR #628, opened 2022-03-18
Upgrade rails 7

catalinailie:


BhupsAL1 (commented):
LGTM

ChrisBAshton (commented):
Thanks @catalinailie - this is looking good.

A few comments below, and also one here: we'll need to add [cookie rotator code](https://github.com/alphagov/content-tagger/pull/1272/commits/3da34d0bd7f289f7401f42397b4abc465ad1ecff) (see https://guides.rubyonrails.org/v7.0/upgrading_ruby_on_rails.html#key-generator-digest-class-changing-to-use-sha256) and then [remove it later](https://github.com/alphagov/content-tagger/pull/1278).

config/environments/test.rb
```diff
@@ -8,13 +8,13 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  config.cache_classes = false
-  config.action_view.cache_template_loading = true
+  # Turn false under Spring and add config.action_view.cache_template_loading = true.
+  config.cache_classes = true
```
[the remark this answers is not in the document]
catalinailie: Yes, you are right. As seen on the rails upgrade guide, this should be `false`. This was set after running `rails app:update` and I missed to remove the change.

config/environments/production.rb
```diff
@@ -54,30 +54,19 @@
...
-  config.action_mailer.notify_settings = {
-    api_key: Rails.application.secrets.notify_api_key,
-  }
-
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation cannot be found).
   config.i18n.fallbacks = true
 
-  # Send deprecation notices to registered listeners.
-  config.active_support.deprecation = :notify
```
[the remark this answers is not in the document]
ChrisBAshton: Ah, this might be fine to leave - `notify` is just a Rails thing, not a GOV.UK Notify thing 😁  https://api.rubyonrails.org/classes/ActiveSupport/Deprecation/Behavior.html

Gemfile.lock
```diff
@@ -410,10 +410,7 @@ GEM
       docile (~> 1.1)
       simplecov-html (~> 0.11)
     simplecov-html (0.12.2)
-    spring (2.1.0)
-    spring-watcher-listen (2.0.1)
-      listen (>= 2.7, < 4.0)
-      spring (>= 1.2, < 3.0)
+    spring (4.0.0)
```
ChrisBAshton, line 413: Could you explain in the commit message why you've updated Spring, and why you've had to remove spring-watcher-listen?

config/application.rb
```diff
@@ -41,5 +41,22 @@ class Application < Rails::Application
     config.after_initialize do
       config.action_mailer.notify_settings = { api_key: Rails.application.secrets.notify_api_key }
     end
+
+    # Rotate SHA1 cookies to SHA256 (the new Rails 7 default)
+    # TODO: Remove this after existing user sessions have been rotated
+    # https://guides.rubyonrails.org/v7.0/upgrading_ruby_on_rails.html#key-generator-digest-class-changing-to-use-sha256
+    Rails.application.config.action_dispatch.cookies_rotations.tap do |cookies|
```
ChrisBAshton, line 48: Thanks for adding this.

The only thing I'm not clear on is when the cookies get rotated - does it happen automatically on deployment, or do we need to do anything (paging @ollietreend!)?

ChrisBAshton (commented):
Thanks @catalinailie - one bit of commit message expansion required, otherwise this looks good to go ✅

BhupsAL1 (commented):
LGTM

config/application.rb
ollietreend: 👋🏻

The purpose of this cookie rotation is to avoid users with existing active sessions from being kicked out when the upgrade is deployed.

It covers this scenario:

1. A user is actively using the application. They logged in when it was running Rails 6, so their session cookie is SHA1'd.
2. The Rails 7 upgrade gets deployed while they're still actively using the app. Rails 7 expects cookies to be SHA256'd, but the user has a cookie created by Rails 6 which is SHA1'd.
3. **With first HTTP request the user makes to the Rails 7 app, this cookie rotation code in `application.rb` automatically converts the cookie to SHA256.** So the session cookie in the user's browser will be overwritten with a SHA256'd equivalent cookie.
4. The user's session is preserved. They continue browsing with the newly converted session cookie.

If the cookie rotation code wasn't there, the Rails 7 app would assume the SHA1'd cookie was corrupted/broken and it'd reset the session. So the user would lose their session state and have to login again, etc.

All new session cookies will be SHA256'd. This rotation is only needed to preserve existing sessions of users who would otherwise lose their session. **And once they've made one single HTTP request to the app, their cookie will have been rotated and this code will no longer be needed.**

So in the Publishing team, we're planning to remove the rotation after ~24 hours. That's long enough for any active sessions to be seamlessly converted. And for any users who haven't accessed the app within that time, their session would have naturally expired anyway, so there would be no need to preserve the session.

## TL;DR

Cookie rotation happens automatically when users make HTTP requests to the app. But this rotation code will be redundant after ~24 hours and should be removed.

ChrisBAshton (approved):
Nice one ⭐

config/application.rb
ChrisBAshton: Thanks for the detailed info, @ollietreend - that makes perfect sense ⭐ 

So @catalinailie you're welcome to raise a `"Revert 'Rotate SHA1 encrypted cookies to SHA256'" commit in a PR after this. We'll deploy #628 through to Production and then wait 24 hours before merging the revert PR 👍

ollietreend: @ChrisBAshton That said, I'm not sure cookie rotation should actually be needed in this app.

Since it's an API app, I'm guessing it's stateless and doesn't actually make use of session cookies. I can't see any evidence of it having a web UI or user logins.

PR #760, opened 2023-03-03
Allow AWS auth methods other than secret key to work.

sengi:
This lets us stop using long-lived credentials, for example by switching to EC2 instance profile creds or pod identity (IRSA in Amazon parlance).

The Fog library kinda reinvents the wheel when it comes to AWS SDK authentication, hence we have to set use_iam_profile ourselves according to whether we're passing an access key. `aws_access_key_id` and `aws_secret_access_key` are required args, even when `use_iam_profile` is true.


sengi:
Brakeman failure is just because this app is still on Ruby 2.7.

catalinailie (approved):

sengi:
Thanks!

PR #769, opened 2023-04-13
Use the default Rails logger for organisation importer

theseanything:
Unclear why this class was using it's own longer - but doesn't seem to do anything special expect use JSON formatting and only output in production. This caused an issue with running the `api_sync:import_organisations` rake task on Kubernetes as the logger was configured to write to a file instead of stdout. (Containers have readonly file permissions).


sihugh (approved):
Amazing!

PR #911, opened 2024-03-11
Fix S3 uploads using EC2 instance profile authn.

sengi:
- Remove Fog; it does too much and it rolls its own instead of using Amazon's client libraries for authentication.
- Use the standard Aws::S3::Client library so that we can use more secure authentication methods such as EC2 instance profile creds or IAM Roles for Service Accounts (IRSA / pod identity). This means we can eliminate long-lived IAM credentials (access keys), which reduces toil and improves security.
- Simplify the tests so that we're no longer mixing up testing that the report generator produced the right output vs. testing that the worker calls the report generator and does stuff with the report.
- Stop trying to healthcheck S3.


AgaDufrat (approved):
LGTM. Good call to remove tests for the CSV generation logic which is tested in the models.

hwrightson:
Hi @sengi @AgaDufrat, since this change was made (14th March I believe), the GOVUK feedback pipeline has been failing to run. Specifically it looks as though all subsequent records from the support API are associated with the 14th March rather than being spread out to their actual date. Would either of you be available to jump on a call and take a look at my issue and see whether there is anything here that might have modified the behaviour? Many thanks, H

sengi:
I agree the timing is interesting. It's hard to imagine a mechanism for this PR to cause such an issue though.

This PR only changes how the application connects to S3, and there are still new files appearing, for example:

```
aws s3 ls s3://govuk-production-support-api-csvs/ |tail
...
2024-04-02 09:59:17       1120 feedex_2024-03-25_2024-03-31_done_apply-register-design.csv
2024-04-02 07:07:21      37450 feedex_2024-03-25_2024-03-31_done_report-a-death-to-government.csv
2024-04-02 12:19:42       3598 feedex_2024-03-26_2024-04-01_done_apply-replace-veterans-badge.csv
```

I think you're best bet is to look into this with one of the app's maintainers, who will have a better knowledge of the data pipeline.

AgaDufrat:
@hwrightson as discussed in another platform the issue described is not related to this change or even this S3 bucket.

PR #1154, opened 2025-04-01
Upgrade to Rails 8.0.2

brucebolt:
Upgrades Rails to version 8.0.2.

[Trello card](https://trello.com/c/NkfCs5ci)


config/environments/production.rb
```diff
@@ -40,29 +30,24 @@
...
 
-  # "info" includes generic and useful information about system operation, but avoids logging too much
-  # information to avoid inadvertent exposure of personally identifiable information (PII). If you
-  # want to log everything, set the level to "debug".
+  # Change to "debug" to log everything (including potentially personally-identifiable information!)
   config.log_level = ENV.fetch("RAILS_LOG_LEVEL", "info")
 
-  # Use a different cache store in production.
+  # Prevent health checks from clogging up the logs.
+  config.silence_healthcheck_path = "/up"
```
yndajas, line 41: **todo:** I think this should be commented out? Was is uncommented by default? If so, maybe we should reword [this](https://docs.publishing.service.gov.uk/manual/guidance-for-upgrades-to-rails-8-0.html#:~:text=Leave%20the%20silencing%20of%20the%20healthcheck%20path%20commented%20out.%20We%20have%20a%20custom%20healthcheck%20path%2C%20and%20don%E2%80%99t%20use%20this%20feature.) to say "Comment out" rather than "Leave [...] commented out"

Gemfile.lock
```diff
@@ -187,7 +187,7 @@ GEM
     http-accept (1.7.0)
     http-cookie (1.0.8)
       domain_name (~> 0.5)
-    httparty (0.22.0)
```
yndajas, line 190: **thought:** should we add this `bundle update` step to the guidance?

yndajas (commented):
Looks like the database schema needs updating

config/environments/production.rb
brucebolt: It was indeed uncommented, so I agree that the wording should be changed to "Comment out".

I'll comment it out in this PR.

brucebolt:
> Looks like the database schema needs updating

Thanks, I've dumped the schema and made the changes in the comments.

yndajas (approved):

PR #1155, opened 2025-04-01
Add GitHub Actions to dependabot

brucebolt:
We are using some outdated versions of GitHub Action dependencies, which is causing `actionlint` to fail on the main branch.

Adding GitHub Actions to dependabot to automatically upgrade those dependencies, so the linting checks will pass.


JonathanHallam (approved):
👍

PR #1219, opened 2025-07-02
[PP-7236] Reject null bytes in problem reports

brucebolt:
We are seeing bots submitting our problem report form with content that contains null bytes. This results in an exception being raise: `ArgumentError string contains null byte (ArgumentError)`.

These characters serve no purpose, so we should reject them. This will remove the Sentry alerts and stop spam that contains these characters.


callumknights (approved):
Nice 👍

PR #1456, opened 2026-07-23
[PP-7021] Bump `govuk_sidekiq` to version 12.0.0

brucebolt:
This includes an upgrade to use Sidekiq 8.

### DIFF
--- a/spec/support/sidekiq.rb
+++ b/spec/support/sidekiq.rb
@@ -1,2 +1,3 @@
 require "govuk_sidekiq/testing"
-require "sidekiq/testing/inline"
+
+Sidekiq.testing!(:inline)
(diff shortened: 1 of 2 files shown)


spec/support/sidekiq.rb
```diff
@@ -1,2 +1,2 @@
 require "govuk_sidekiq/testing"
-require "sidekiq/testing/inline"
+Sidekiq.testing!(:inline)
```
yndajas, line 3: **nitpick:** new line after the `require`? I guess Rubocop doesn't care though

yndajas (dismissed):

spec/support/sidekiq.rb
brucebolt: Good idea.  I've added the new line.

yndajas (approved):

